PR #1, opened 2023-12-07
Add nexus config and artifacts

nambrot:


deployments/ism/nexus/config.yaml
```diff
@@ -0,0 +1,71 @@
+# yaml-language-server: $schema=./core-deployments.schema.json
+---
+description: ISM config for Nexus
+timestamp: 2023-09-23T16:00:00.000Z
```
jmrossy, line 4: Is this timestamp right?

deployments/warp_routes/nexus_arbitrum/config.yaml
```diff
@@ -0,0 +1,28 @@
+a# yaml-language-server: $schema=./core-deployments.schema.json
```
jmrossy, line 1: typo leading `a` here

jmrossy:
No objections to the schema here though I'm confused why you included just the warp + ism configs/artifacts and not the core artifacts for these chains

nambrot:
The main goal was to have a reference for the warp routes because that was blocking folks. I suspect core artifacts can come later

deployments/ism/nexus/config.yaml
nambrot: No, but I believe currently unused? I just had it copy-pasted

PR #2, opened 2024-02-03
feat:switch toplevel keys to chainNames

aroralanuk:
the monorepo has chainName as the toplevel keys so to maintain consistency adjusted the warp route config to match it.

### DIFF
--- a/deployments/warp-deployments.yaml
+++ b/deployments/warp-deployments.yaml
@@ -4,39 +4,27 @@
 ---
 description: Hyperlane Warp Route artifacts
 timestamp: 2023-09-23T16:00:00.000Z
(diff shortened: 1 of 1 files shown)


deployments/warp-deployments.yaml
```diff
@@ -4,39 +4,27 @@
 ---
 description: Hyperlane Warp Route artifacts
 timestamp: 2023-09-23T16:00:00.000Z
+deployer: Abacus Works (Hyperlane)
```
jmrossy, line 7: The deployer and timestamp are under the `data` list because this file was intended to contain *all* warp deployments from anyone.

jmrossy:
Outdated

PR #4, opened 2024-03-19
Add Infra warp artifacts

yorhodes:
From https://github.com/hyperlane-xyz/hyperlane-monorepo/tree/main/typescript/infra/config/environments/mainnet3/warp

Adapts to schema from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3379


deployments/warp_routes/README.md
```diff
@@ -0,0 +1,15 @@
+Directory structure is as follows:
+
+```
+<ASSET_SYMBOL>/<ORIGIN_CHAIN>/<DESTINATION_CHAIN>/addresses.json
+```
+
+Addresses are keyed by their token router type:
+
+```typescript
+enum TokenType {
```
aroralanuk, line 10: tokenType can be fastSynthetic, nativeScaled, etc if we're going by the sdk

deployments/warp_routes/README.md
```diff
@@ -0,0 +1,15 @@
...
+  synthetic,
+  collateral,
+  native,
+}
+```
```
aroralanuk, line 15: also if it's possible, maybe we can directly read from the monorepo similar to the docs

aroralanuk (commented):
My assumption was we were using warp core style yaml artifacts in registry but I'll leave it to @jmrossy

jmrossy:
+1 for Kunal's suggestion of using the WarpCoreConfig shape (aka the warp ui config shape), which is more descriptive, multi-protocol, and has the info one would need to use a warp route with our tooling

yorhodes:
> +1 for Kunal's suggestion of using the WarpCoreConfig shape (aka the warp ui config shape), which is more descriptive, multi-protocol, and has the info one would need to use a warp route with our tooling

the artifacts here are also multi protocol
I included these because this is whats generated by the deployments, the fields on the `WarpCoreConfig` shape can be derived from this address (I realize we want to do this at build time in the warp UI)
imo someone shouldnt have to submit a PR to this repo to change eg the hook config of their warp route

jmrossy:
> I included these because this is whats generated by the deployments

The deployments generate both these and the warp core config, so what's the advantage of using this one?

> the fields on the WarpCoreConfig shape can be derived from this address

I don't think they can. An address is not sufficient is some cases, like to differentiate from a Sealevel SPL vs SPL 2022 token which is necessary information to query the program.

> I realize we want to do this at build time in the warp UI

I actually do not want to do this at build time. I feel pretty strongly about this. Build time is the worst of both worlds. It means we have to manage the complexity of querying tokens and mapping the routes **and** still need to trigger de-redeployments if the chain state ever changes.  

>  imo someone shouldnt have to submit a PR to this repo to change eg the hook config of their warp route

Why would they need to do that if we use the WarpCore config shape here instead of this minimal representation.

--

@yorhodes I guess my main criticism here is that we **just recently** unified the tooling around the WarpCore config shape and now this deviates away from it. I know you didn't fully agree with the verbosity of that shape but I still feel strongly it was the way to go, mainly because we did try this minimal config approach before and had to replace it.

yorhodes:
> @yorhodes I guess my main criticism here is that we just recently unified the tooling around the WarpCore config shape and now this deviates away from it. I know you didn't fully agree with the verbosity of that shape but I still believe it was the right decision. Keep in mind, we did try this minimal config approach in the warp UI before and ultimately had to replace it.

for posterity, in the short term we will add the warp core configs alongside the address artifacts

PR #3, opened 2024-03-19
Add SDK chain metadata and address consts

yorhodes:
generated from
```typescript
for (const [chain, metadata] of Object.entries(chainMetadata)) {
  fs.mkdirSync(`chains/${chain}`, { recursive: true });
  fs.writeFileSync(`chains/${chain}/metadata.json`, JSON.stringify(metadata, null, 2));
  // @ts-ignore
  const addresses: string | undefined = mainnet[chain] ?? testnet[chain];
  if (addresses) {
    fs.writeFileSync(`chains/${chain}/addresses.json`, JSON.stringify(addresses, null, 2));
  }
}
```

TODO: remove hooks and ISMs (should be derivable)

PR #5, opened 2024-04-07
SDK const migration

jmrossy:
- Finish migration of SDK consts started in #3 
- Convert json chain files to yaml
- Create build script to transform data
- Set up CI job and mocha tests
- Productize package.json file
- Update Readme
- Lock yarn version like in monorepo


chains/polygon/addresses.yaml
```diff
@@ -0,0 +1,21 @@
+aggregationHook: "0x34dAb05650Cf590088bA18aF9d597f3e081bCc47"
```
yorhodes, line 1: should we omit the quotes?

scripts/migrate-sdk-consts.js
yorhodes, line 1: nice
should we remove this or keep jic?

scripts/update-chain-schema.js
```diff
@@ -0,0 +1,7 @@
+import fs from 'fs'
+import { zodToJsonSchema } from "zod-to-json-schema";
+import {ChainMetadataSchemaObject} from "@hyperlane-xyz/sdk";
+
+const schema = zodToJsonSchema(ChainMetadataSchemaObject, "hyperlaneChainMetadata");
```
yorhodes, line 5: this is cool!

chains/schema
yorhodes, line 1: should we git ignore this if its generated?

scripts/build.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function genJsExport(data, exportName) {
+  return `export const ${exportName} = ${JSON.stringify(data, null, 2)};`;
+}
+
+let chainMetadata = {};
+let chainAddresses = {};
+
+console.log('Parsing and copying chain data');
+for (const file of fs.readdirSync('./chains')) {
```
yorhodes, line 14: what about the core.ts and schema files?

scripts/build.ts
```diff
@@ -0,0 +1,76 @@
...
+  const inDirPath = `./chains/${file}`;
+  const assetOutPath = `./dist/chains/${file}`;
+  // generated ts files to to the tmp dir so they can be compiled along with other generated code
+  const tsOutPath = `./tmp/chains/${file}`;
+  const stat = fs.statSync(`${inDirPath}`);
+  if (!stat.isDirectory()) continue;
```
yorhodes, line 20: ah I see

scripts/build.ts
```diff
@@ -0,0 +1,76 @@
+import { pick } from '@hyperlane-xyz/utils';
+import fs from 'fs';
+import { parse } from 'yaml';
+import { CoreChains } from '../chains/core';
+
+function genJsExport(data, exportName) {
+  return `export const ${exportName} = ${JSON.stringify(data, null, 2)};`;
+}
```
yorhodes, line 8: is this a common pattern? kinda surprised that we are exporting non-static assets tbh

yorhodes: how useful is this?

scripts/build.ts
yorhodes, line 1: slightly worried this is fragile but love the backward compatibility

test/chains.test.ts
yorhodes, line 1: nice!

yorhodes: could we adapt this to use the bundled exports and skip the readAllFiles util?

test/chains.test.ts
```diff
@@ -0,0 +1,24 @@
+import { ChainMetadataSchema } from '@hyperlane-xyz/sdk';
+import { z } from 'zod';
+import { readAllChainYamlFiles } from './utils';
+
+describe('Chain metadata', () => {
+  it('All chains have valid metadata', () => {
+    const result = readAllChainYamlFiles('metadata.yaml');
+    for (const [chain, metadata] of Object.entries(result)) {
+      console.log('Validating metadata for', chain);
+      ChainMetadataSchema.parse(metadata);
```
yorhodes, line 10: nit but slightly prefer separate 
```it(`${chain} metadata is valid`, () =>```

yorhodes (approved):
this looks great

chains/polygon/addresses.yaml
jmrossy: I believe yaml may consider it a number. I didn't put them in explicitly, this is just what the yaml lib spit out

scripts/migrate-sdk-consts.js
jmrossy: I'd recommend we keep jib for a week or two. I'll take the AI to remove it when things are settled.

scripts/build.ts
jmrossy: core.ts gets copied over into index.ts which then gets compiled by ts.
The schema I forgot, I'll get that included too

scripts/build.ts
jmrossy: I feel good about it personally. It's simple stuff even if we need to overhaul

test/chains.test.ts
jmrossy: Oh maybe, I'll give it a try now

test/chains.test.ts
jmrossy: You got it 👍

test/chains.test.ts
jmrossy: Done, good idea

PR #9, opened 2024-04-10
Add metadata health tests and improve export types

jmrossy:


test/health/explorer-health.test.ts
```diff
@@ -0,0 +1,81 @@
+import { ChainMetadata } from '@hyperlane-xyz/sdk';
```
jmrossy, line 1: Not much new code in the next 3 files, mostly ported from metadata check in the SDK
I considered keeping it there but it's only used in this one place.

yorhodes (commented):
thanks!

yorhodes (approved):

PR #11, opened 2024-04-16
Create Registry utility classes

jmrossy:
Port new registry tooling from SDK
First introduced in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3605


jmrossy:
New code as approved in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3605 so I'm going ahead with merge

PR #13, opened 2024-04-17
Implement LocalRegistry

jmrossy:
- Create new LocalRegistry class
- Add new export entry for local-related things
- Refactor unit tests to cover both registry types


src/registry/LocalRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+  constructor(options: LocalRegistryOptions) {
+    super({ logger: options.logger });
+    this.uri = options.uri;
+  }
+
+  listRegistryContent(): RegistryContent {
+    if (this.listContentCache) return this.listContentCache;
+
+    const chainFileList = this.listFiles(path.join(this.uri, this.getChainsPath()));
+    const chains: ChainMap<ChainFiles> = {};
```
jmrossy, line 40: There's some code duplication between this file and GithubRegistry but importantly, all the methods here are sync, which makes code reuse tricky. Given that it's <30 dupe lines total, I'm just going with this for now. We can revisit if more duplication arises when adding the write-related methods

test/unit/registry.test.ts
```diff
@@ -0,0 +1,52 @@
+import { expect } from 'chai';
+
+import { GithubRegistry } from '../../src/registry/GithubRegistry.js';
+import { LocalRegistry } from '../../src/registry/LocalRegistry.js';
+
+describe('Registry utilities', () => {
```
jmrossy, line 6: Not really new, this is a modified `GithubRegistry.test.ts` but I guess git didn't make the connection

src/registry/LocalRegistry.ts
yorhodes: could you use your MaybePromise thing somehow?

src/registry/LocalRegistry.ts
```diff
@@ -0,0 +1,102 @@
...
+    const addresses = this.getAddresses();
+    if (!addresses[chainName])
+      throw new Error(`Addresses not found in registry for chain: ${chainName}`);
+    return addresses[chainName];
+  }
+
+  protected listFiles(dirPath: string): string[] {
+    let entries = fs.readdirSync(dirPath, { withFileTypes: true });
+
+    let filePaths = entries.map((entry) => {
```
yorhodes, line 95: nit: const?

src/utils.ts
yorhodes, line 1: rm?

src/registry/LocalRegistry.ts
yorhodes: having all of the per/multi chain stuff be reused might be nice

yorhodes (approved):

src/registry/LocalRegistry.ts
jmrossy: I've set up eslint to catch these, thanks

PR #15, opened 2024-04-18
Implement methods to adding/removing chains to a LocalRegistry

jmrossy:


src/registry/BaseRegistry.ts
```diff
@@ -36,4 +36,11 @@ export abstract class BaseRegistry implements IRegistry {
   abstract getChainMetadata(chainName: ChainName): MaybePromise<ChainMetadata | null>;
   abstract getAddresses(): MaybePromise<ChainMap<ChainAddresses>>;
   abstract getChainAddresses(chainName: ChainName): MaybePromise<ChainAddresses | null>;
+  abstract addChains(
```
yorhodes, line 39: should we have single chain interface to mirror the `get`?

src/registry/LocalRegistry.ts
```diff
@@ -89,6 +91,65 @@ export class LocalRegistry extends BaseRegistry implements IRegistry {
...
   }
 
+  addChains(
+    chains: Array<{ chainName: ChainName; metadata?: ChainMetadata; addresses?: ChainAddresses }>,
+  ): void {
+    const currentChains = this.listRegistryContent().chains;
+    for (const chain of chains) {
+      if (currentChains[chain.chainName])
+        throw new Error(`Chain ${chain.chainName} already exists in registry`);
+    }
```
yorhodes, line 101: should we throw or merge here?

src/registry/LocalRegistry.ts
```diff
@@ -99,4 +160,37 @@ export class LocalRegistry extends BaseRegistry implements IRegistry {
...
+  protected createChainFile(
+    chainName: ChainName,
+    fileName: keyof ChainFiles,
+    data: any,
+    cache: ChainMap<any>,
+    prefix?: string,
+  ) {
+    const filePath = path.join(this.uri, this.getChainsPath(), chainName, `${fileName}.yaml`);
+    this.listContentCache!.chains[chainName] ||= {};
+    this.listContentCache!.chains[chainName][fileName] = filePath;
```
yorhodes, line 182: why is this definitely defined here?

src/registry/LocalRegistry.ts
```diff
@@ -99,4 +160,37 @@ export class LocalRegistry extends BaseRegistry implements IRegistry {
...
+    cache[chainName] = data;
+    this.createFile({ filePath, data: toYamlString(data, prefix) });
+  }
+
+  protected removeFiles(filePaths: string[]): void {
+    for (const filePath of filePaths) {
+      fs.unlinkSync(filePath);
```
yorhodes, line 189: is this a lock?

src/registry/LocalRegistry.ts
```diff
@@ -89,6 +91,65 @@ export class LocalRegistry extends BaseRegistry implements IRegistry {
...
+        this.createChainFile(
+          chainName,
+          'metadata',
+          chain.metadata,
+          this.metadataCache!,
+          CHAIN_SCHEMA_REF,
+        );
+      }
+      if (chain.addresses) {
+        this.createChainFile(chainName, 'addresses', chain.addresses, this.addressCache!);
```
yorhodes, line 133: would we ever add an addresses schema later?

src/registry/LocalRegistry.ts
```diff
@@ -89,6 +91,65 @@ export class LocalRegistry extends BaseRegistry implements IRegistry {
...
+    this.getAddresses();
+
+    for (const chain of chains) {
+      const chainName = chain.chainName;
+      if (chain.metadata) {
+        this.createChainFile(
+          chainName,
+          'metadata',
+          chain.metadata,
+          this.metadataCache!,
```
yorhodes, line 128: slightly dont like this block, could we have the cache be returned in a var or have createChainFile read from the cache?

src/registry/GithubRegistry.ts
```diff
@@ -123,6 +123,20 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
...
+    throw new Error('TODO: Implement');
+  }
+  async updateChains(
+    _chains: Array<{ chainName: ChainName; metadata?: ChainMetadata; addresses?: ChainAddresses }>,
+  ): Promise<void> {
+    throw new Error('TODO: Implement');
+  }
+  async removeChains(_chains: Array<ChainName>): Promise<void> {
+    throw new Error('TODO: Implement');
+  }
```
yorhodes, line 138: honestly not sure if we will ever implement this now that I think about it

yorhodes: we probably want to do the syncing all at once

yorhodes (approved):
lgtm, not sure if github registry should have the same `write` interface

src/registry/LocalRegistry.ts
jmrossy: No, this is how you delete a file with node's fs lib. Strange I know.

src/registry/LocalRegistry.ts
jmrossy: IMO people should use `updateChain` if they want add-or-merge behavior

src/registry/LocalRegistry.ts
jmrossy: We can if you'd like, we just need a set list of contract names. Gonna punt on it for now

src/registry/LocalRegistry.ts
jmrossy: Reworked it a bit, thanks for the suggestion

PR #21, opened 2024-04-25
Warp Route add support

jmrossy:
- Convert warp route artifacts to yaml
- Add addWarpRoute method to IRegistry


yorhodes (changes requested):
I still think that the warp core config is much more detailed/confusing than necessary for a minimal artifact schema and the addresses format is much easier to work with and would prefer not to remove these
I also do not expect every consumer of the registry to use the NPM package

yorhodes (approved):

PR #27, opened 2024-05-10
feat: add holesky addresses

aroralanuk:
#### Other notes

ProxyAdmin with owner: 0xfaD1C94469700833717Fa8a3017278BC1cA8031C

### Backward compatibility

Yes

### Testing

Have any new metadata configs and deployment addresses been used with any Hyperlane tooling, such as the CLI?

No

### DIFF
--- a/.changeset/nice-bottles-unite.md
+++ b/.changeset/nice-bottles-unite.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Add Holesky AVS deployment addresses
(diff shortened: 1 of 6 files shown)


paulbalaji (commented):
would be good to add a simple README at the root of avs dir, a la https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/deployments/warp_routes/README.md

paulbalaji (approved):
would also be good to add a simple README at the root of avs dir, a la https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/deployments/warp_routes/README.md

PR #32, opened 2024-05-16
Migrate MergedRegistry from monorepo

jmrossy:
### Description

- Migrate MergedRegistry from the monorepo's CLI package to here
- Add required objMerge utility

### Backward compatibility

Yes

### Testing

New unit test coverage


src/utils.ts
```diff
@@ -19,3 +19,33 @@ export async function concurrentMap<A, B>(
...
   return res;
 }
+
+export function isObject(item: any) {
+  return item && typeof item === 'object' && !Array.isArray(item);
+}
+
+// Recursively merges b into a
+// Where there are conflicts, b takes priority over a
+export function objMerge(a: Record<string, any>, b: Record<string, any>, max_depth = 10): any {
```
jmrossy, line 29: The reason I've duplicated this utils code here is to avoid taking a dependency on `@hyperlane-xyz/utils` here. Three reasons:
1) Because of the crypto libs, utils is big. People looking for just addresses don't want it.
2) Monorepo packages use the registry lib. If registry also has a runtime dep on a monorepo package, things can get weird.
3) Given that this package is versioned independently of the monorepo, a util and/or sdk runtime dep here would likely lead to duplicate versions getting bundled into apps eventually

jmrossy: Obviously there's some threshold for how much duplication we can tolerate before it's maybe worth the downsides but I think we're still below that threshold.

src/registry/MergedRegistry.ts
yorhodes, line 1: personally think it would be nice for the `BaseRegistry` to have a `merge(other: IRegistry)` so users dont need to be aware of this class nor instantiate a new registry

yorhodes (approved):

src/registry/MergedRegistry.ts
jmrossy: Done in https://github.com/hyperlane-xyz/hyperlane-registry/pull/33/

PR #33, opened 2024-05-16
Add PartialRegistry class and merge method

jmrossy:
### Description

Builds on #32 

### Backward compatibility

Yes

### Testing

New unit test coverage


src/registry/MergedRegistry.ts
```diff
@@ -8,17 +8,28 @@ import { IRegistry, RegistryContent, RegistryType } from './IRegistry.js';
 
 export interface MergedRegistryOptions {
   registries: Array<IRegistry>;
+  chainMetadataOverrides?: ChainMap<ChainMetadata>;
```
tkporter, line 11: should this be `ChainMap<Partial<ChainMetadata>>` as well?

jmrossy: yes!

src/registry/MergedRegistry.ts
```diff
@@ -3,22 +3,30 @@ import type { Logger } from 'pino';
...
 }
 
-export class MergedRegistry extends BaseRegistry implements IRegistry {
+/**
+ * A registry that accepts multiple sub-registries.
+ * Read methods are performed on all sub-registries and the results are merged.
+ * Write methods are performed on all sub-registries.
+ * Can be created manually or by calling `.merge()` on an existing registry.
+ */
+export class MergedRegistry implements IRegistry {
```
jmrossy, line 19: Note, this class can't extend BaseRegistry anymore because BaseRegistry uses it in `merge` but it's probably fine because this class should remain simple

src/registry/MergedRegistry.ts
```diff
@@ -51,6 +59,11 @@ export class MergedRegistry extends BaseRegistry implements IRegistry {
     return (await this.getAddresses())[chainName] || null;
   }
 
+  async getChainLogoUri(chainName: ChainName): Promise<string | null> {
+    const results = await this.multiRegistryRead((r) => r.getChainLogoUri(chainName));
+    return results.filter((uri) => uri !== null)[0] || null;
```
tkporter, line 64: nit: definitely don't feels strongly but could use `.find` which short circuits upon a truthy value

src/registry/PartialRegistry.ts
```diff
@@ -0,0 +1,74 @@
...
+      chains[c].addresses = PARTIAL_URI_PLACEHOLDER;
+    });
+    return {
+      chains,
+      deployments: {},
+    };
+  }
+
+  getMetadata(): ChainMap<ChainMetadata> {
+    return this.chainMetadata as ChainMap<ChainMetadata>;
```
tkporter, line 49: I assume it's intentional to accept that it'd be possible for someone to end up with a PartialRegistry treated as if it were a full registry, & that dealing with fully accurate types wasn't worth the hassle?

tkporter (approved):

src/registry/MergedRegistry.ts
jmrossy: Ah yes should have thought of that thanks

src/registry/PartialRegistry.ts
jmrossy: Exactly. It's easy to use this wrong but the I don't want to weaken the types everywhere else just to accommodate this thing that we'll probably be the only ones to use

PR #43, opened 2024-05-23
Warp route config read support

jmrossy:
### Description

Add support for reading warp route configs from registries
Required for https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3692

### Backward compatibility

Technically yes but marking changeset as major to compensate for incorrect version bump in last version. See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3821

### Testing

New unit tests


scripts/build.ts
```diff
@@ -20,102 +25,172 @@ function genChainMetadataMapExport(data, exportName) {
 ${genJsExport(data, exportName)} as ChainMap<ChainMetadata>`;
 }
 
-console.log('Preparing tmp directory');
```
jmrossy, line 23: The diff in this file is a lot less than it seems. I just wrapped the chunks in functions. Use the PR "Hide Whitspace" setting for a better view.

yorhodes: thanks for the tip!

scripts/build.ts
```diff
@@ -20,102 +25,172 @@ function genChainMetadataMapExport(data, exportName) {
...
+    // ts files go to the tmp dir so they can be compiled along with other generated code
+    const tsOutPath = `./tmp/${warpPathBase}/${warpDir}`;
+    const stat = fs.statSync(`${inDirPath}`);
+    if (!stat.isDirectory()) continue;
 
-  // Copy the logo file
-  fs.copyFileSync(`${inDirPath}/logo.svg`, `${assetOutPath}/logo.svg`);
+    // Inner loop for individual warp route configs
+    for (const warpFile of fs.readdirSync(inDirPath)) {
+      if (!warpFile.endsWith('config.yaml')) continue;
```
yorhodes, line 92: is this really necessary? I guess this assumes the CLI generated this

scripts/build.ts
```diff
@@ -20,102 +25,172 @@ function genChainMetadataMapExport(data, exportName) {
...
+      const [warpFileName] = warpFile.split('.');
+      const config = parse(fs.readFileSync(`${inDirPath}/${warpFile}`, 'utf8'));
+      const id = warpRouteConfigToId(config);
```
yorhodes, line 95: isnt this pretty similar to the filepath? whats the benefit here

src/registry/FileSystemRegistry.ts
```diff
@@ -83,6 +89,17 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
...
+  addWarpRoute(config: WarpCoreConfig): void {
+    let { configPath, addressesPath } = this.getWarpRoutesArtifactPaths(config);
+
+    configPath = path.join(this.uri, configPath);
+    this.createFile({ filePath: configPath, data: toYamlString(config, SCHEMA_REF) });
+
+    addressesPath = path.join(this.uri, addressesPath);
+    const addresses = warpConfigToWarpAddresses(config);
+    this.createFile({ filePath: addressesPath, data: toYamlString(addresses) });
+  }
```
yorhodes, line 101: why move?

src/registry/FileSystemRegistry.ts
```diff
@@ -166,4 +168,15 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
       fs.rmdirSync(parentDir);
     }
   }
+
+  protected getWarpRoutesForIds(ids: WarpRouteId[]): WarpCoreConfig[] {
```
yorhodes, line 172: seems like unnecessary indirection from IDs <> filepaths

src/registry/GithubRegistry.ts
```diff
@@ -108,30 +120,37 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
...
 
+  async getWarpRoute(routeId: string): Promise<WarpCoreConfig | null> {
+    const repoContents = await this.listRegistryContent();
+    const routeConfigUrl = repoContents.deployments.warpRoutes[routeId];
+    if (!routeConfigUrl) return null;
+    return this.fetchYamlFile(routeConfigUrl);
+  }
+
+  async getWarpRoutes(filter?: WarpRouteFilterParams): Promise<WarpRouteConfigMap> {
+    const warpRoutes = (await this.listRegistryContent()).deployments.warpRoutes;
```
yorhodes, line 147: does `listRegistryContent` cache?

src/registry/IRegistry.ts
```diff
@@ -10,10 +10,22 @@ export interface ChainFiles {
 export interface RegistryContent {
   chains: ChainMap<ChainFiles>;
   deployments: {
-    // TODO define deployment artifact shape here
+    // Map of warp route ID to config URI
+    warpRoutes: Record<WarpRouteId, string>;
```
yorhodes, line 14: this seems like a departure from the ChainFiles?

src/registry/IRegistry.ts
```diff
@@ -10,10 +10,22 @@ export interface ChainFiles {
...
 
+export interface UpdateChainParams {
+  chainName: ChainName;
+  metadata?: ChainMetadata;
+  addresses?: ChainAddresses;
+}
+
+export interface WarpRouteFilterParams {
+  symbol?: string;
+  chainName?: ChainName;
```
yorhodes, line 26: nit: we use chain in most other places, why chainName now?

src/registry/MergedRegistry.ts
```diff
@@ -93,6 +93,16 @@ export class MergedRegistry implements IRegistry {
     );
   }
 
+  async getWarpRoute(id: WarpRouteId): Promise<WarpCoreConfig | null> {
+    const results = await this.multiRegistryRead((r) => r.getWarpRoute(id));
+    return results.find((r) => !!r) || null;
+  }
+
+  async getWarpRoutes(filter?: WarpRouteFilterParams): Promise<WarpRouteConfigMap> {
```
yorhodes, line 101: what if I want to specify multiple symbols or chains?

src/types.ts
```diff
@@ -1,7 +1,19 @@
...
 // https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#the-awaited-type-and-promise-improvements
 export type MaybePromise<T> = T | Promise<T> | PromiseLike<T>;
 
 export const ChainAddressesSchema = z.record(z.string());
 export type ChainAddresses = z.infer<typeof ChainAddressesSchema>;
+
+export type WarpRouteId = string;
+export type WarpRouteConfigMap = Record<WarpRouteId, WarpCoreConfig>;
+
+export type RecursivePartial<T> = {
```
yorhodes, line 13: woah
```ts
type DeepPartial<T> = T extends object ? {
    [P in keyof T]?: DeepPartial<T[P]>;
} : T;
```
does this work too? from https://stackoverflow.com/questions/61132262/typescript-deep-partial

src/registry/warp-utils.ts
```diff
@@ -19,3 +24,60 @@ function getWarpAddressKey(standard: TokenStandard): string | null {
...
+  const [_, tokenSymbol, chains] = matches;
+  return createWarpRouteConfigId(tokenSymbol, chains.split('-'));
+}
+
+/**
+ * Gets a warp route ID from a warp route config.
+ */
+export function warpRouteConfigToId(config: WarpCoreConfig): WarpRouteId {
+  if (!config?.tokens?.length) throw new Error('Cannot generate ID for empty warp config');
+  const tokenSymbol = config.tokens[0].symbol;
```
yorhodes, line 46: what if there are multiple token symbols?

src/registry/PartialRegistry.ts
```diff
@@ -12,21 +13,24 @@ const PARTIAL_URI_PLACEHOLDER = '__partial_registry__';
  * Useful for merging with other registries force overrides of subsets of data.
  */
 export interface PartialRegistryOptions {
-  chainMetadata?: ChainMap<Partial<ChainMetadata>>;
-  chainAddresses?: ChainMap<Partial<ChainAddresses>>;
+  chainMetadata?: ChainMap<RecursivePartial<ChainMetadata>>;
+  chainAddresses?: ChainMap<RecursivePartial<ChainAddresses>>;
+  warpRoutes?: Array<RecursivePartial<WarpCoreConfig>>;
```
yorhodes, line 18: no way we can just make this type `Partial<RegistryOptions>`?

yorhodes (commented):
a bit concerned that the user needs to understand these new warp IDs to use these interfaces
I expected this to be an opaque implementation detail

scripts/build.ts
jmrossy: It's so we skip the `-addresses.yaml` file that lives next to this

scripts/build.ts
jmrossy: It is similar yes. The id could have been a file path but:
1. The file paths have some extra stuff in it (`-config.yaml`)
2. After having to work around slashes in cosmos IBC token denoms, I've become weary of using `/` as a separator symbol in our IDs

src/registry/FileSystemRegistry.ts
jmrossy: To have this public above the protected methods. I shouldn't have put it below before

src/registry/GithubRegistry.ts
jmrossy: Yes

src/registry/IRegistry.ts
jmrossy: How so?

src/registry/MergedRegistry.ts
jmrossy: Call again I guess? I'm open to alternative filter designs but didn't want to over-engineer this. 
One though could be a list of filters but some people may expect them the intersection of the filter results, not the union

src/registry/PartialRegistry.ts
jmrossy: It could maybe be a `RecursivePartial` of these but does it matter?

src/registry/warp-utils.ts
jmrossy: Considered that, this would just take the first. Supporting different symbols in warp configs would imply our current symbol-based file organization would need to change. Decided not to pull that thread atm

src/types.ts
jmrossy: Not sure that would handle arrays in objects - like `tokens` in a warp config - but I'll try. If it works that's definitely nicer!

src/registry/FileSystemRegistry.ts
jmrossy: See https://github.com/hyperlane-xyz/hyperlane-registry/pull/43#discussion_r1619116809

src/types.ts
yorhodes: also not sure
I feel like Array might extend object?

scripts/build.ts
yorhodes: oh hmm

scripts/build.ts
jmrossy: Resolved offline, we'll keep these ID utilities but change the format a bit from `USDT|chain1-chain2` to `USDT/chain1-chain2`

src/registry/MergedRegistry.ts
jmrossy: Added info to the function comment header to clarify

src/registry/warp-utils.ts
jmrossy: Added info to the fn comment header to clarify

src/types.ts
jmrossy: Works!

src/registry/IRegistry.ts
yorhodes: those are just filepaths iiuc

src/registry/MergedRegistry.ts
yorhodes: okay yeah no need to complicate now

yorhodes (approved):

src/registry/IRegistry.ts
jmrossy: They're a map of file type to file URI. These are similar, except it's ID to URI. I'll add a comment above `chains` to clarify

PR #46, opened 2024-05-27
Add token logos

jmrossy:
### Description

- Add svg files for token logos
- Define logoURI field in warp configs

### Backward compatibility

Yes


deployments/warp_routes/ECLIP/logo.svg
jmrossy, line 1: Note, this is a bit different than the current ECLIP logo we use on Nexus because this is the only logo version the Eclipse project seems to have that's an SVG.

deployments/warp_routes/ETH/ethereum-viction-config.yaml
yorhodes, line 1: hmm, feels like this violates good version control
I guess the benefit is we don't have to add any logic to the NPM bundling?
would prefer if these were relative to have consistent versioning

yorhodes (approved):
please see my note about versioning

yorhodes (approved):

PR #56, opened 2024-06-08
Neutron warp route configs fixes and tests

jmrossy:
### Description

- Fix warp route file names not matching IDs 
- Add missing token def for neutron-manta route
- Remove dupe chain names in warp route IDs
- Add unit test to validate all warp configs

### Backward compatibility

Yes

### Testing

Added new unit tests


test/unit/warp-routes.test.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from 'chai';
+
+import { MultiProtocolProvider, WarpCore } from '@hyperlane-xyz/sdk';
+import { FileSystemRegistry } from '../../src/registry/FileSystemRegistry.js';
+
+describe('Warp Route Configs', () => {
```
jmrossy, line 6: This will catch any invalid warp configs going forward. The reason for the mistake was that previously all the configs were in one big nexus bucket, meaning they could reference each other easily. Now they're broken apart by symbol and chains

paulbalaji (approved):

yorhodes (approved):

PR #58, opened 2024-06-11
[WIP] feat(registry): expose getWarpArtifactsPaths

nbayindirli:

### DIFF
--- a/src/registry/BaseRegistry.ts
+++ b/src/registry/BaseRegistry.ts
@@ -28,7 +28,7 @@ export abstract class BaseRegistry implements IRegistry {
     return 'chains';
   }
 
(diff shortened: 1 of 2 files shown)


src/registry/BaseRegistry.ts
```diff
@@ -28,7 +28,7 @@ export abstract class BaseRegistry implements IRegistry {
     return 'chains';
   }
 
-  protected getWarpArtifactsPaths({ tokens }: WarpCoreConfig) {
+  getWarpArtifactsPaths({ tokens }: WarpCoreConfig) {
```
jmrossy, line 31: What's motivating this change? If we want something public, we should make all the *Path methods public

PR #88, opened 2024-07-03
feat: add Arthera Mainnet and LESS Warp Route

bertux:
### Backward compatibility

Yes

### Testing

The new metadata configs and deployment addresses have been generated by Hyperlane CLI.


paulbalaji:
please run `yarn prettier`

bertux:
Thanks @paulbalaji for spotting these small fixes, I've accepted commited the suggestions and run prettier.
But in the same way than for the missing of alphabetical order, I don't remember to have added this extra line at the top which was then removed by prettier so it may be caused by empty prefix when the CLI has written these files.
I will check later by trying to add other testnet for more tests.

paulbalaji (approved):

Issue #42, opened 2024-05-22
Add `deployerName` field to registry


avious00:
As a user or developer, I want to know whom the warp route is deployed by so I can understand my trust assumptions. Also has implications for which routes show up in Nexus.

As part of this issue, figure out where this field fits into the registry schema, keeping in mind we may have more sophisticated verification methods in the future, including ENS resolution.

jmrossy:
## Proposal 1

Add a new `deployer` field with `name` and `contact` sub-fields to the chain addresses, warp config, and AVS files.
```
+ Keeps 'owner' info close to artifacts
+ Relatively simple
- Unclear how to handle chains without addresses (e.g. Solana)
- Not very granular (e.g. diff contracts have diff owners)
```

### 1A

Add `deployer` as an optional new field directly at the artifact roots
```
+ Additive so backwards compatible
- Feels out of place among chain contract->address map
```

### 1B

Add `deployer` to the warp + avs artifact roots but restructure the chain address map
```
+ Better end result
- More work than 1A
- Breaking change
```

## Proposal 2

Create new `deployer.yaml` or `owner.yaml` files along-side existing artifacts
```
+ Additive so backwards compatible
+ Probably more flexible for future changes
- Doesn't fit well with warp route structure (grouped by symbol)
- More work than Prop 1A
```

## Other alternatives

1. Add info to chain metadata instead of address artifacts
2. Create a map of address -> name in the registry
3. Use a 3rd party naming protocol like ENS

cmcewen:
@jmrossy what's your personal recommendation?

jmrossy:
@cmcewen Tbh I don't love any of my solutions but I'd lean towards 1B, which is strictly better than 1A but more work and avoids the extra latency more file fetching would add in prop 2

avious00:
@jmrossy sharing from standup that @cmcewen isn't too opinionated, so ultimately your call

jmrossy:
Withdrawing my support for 1A or 1B as it would require an awkward request of information from deployers in the CLI while deploying.

Spoke briefly with Yorke who made a good argument that this info should probably live with other human-defined info in the ChainMetadata instead of with the machine-generated address artifacts. I'll go with that but it will require an SDK change

PR #89, opened 2024-07-04
feat: add `deployer` metadata for all chains

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-registry/issues/42

- updates to latest SDK 4.1.0 which allows addition of `deployer` details to chain metadata
	- update the schema with latest changes
- add `deployer` details for all chains, both AW and PI
- add test to check presence of `deployer` field


paulbalaji:
@avious00 to review deployer details for PI chains

avious00:
@paulbalaji could you change Michalis to Toucan Protocol https://toucan.earth/

and just spelling Merkly as Merkly Labs

otherwise LGTM thanks for getting this out 🙏

chains/basesepolia/metadata.yaml
```diff
@@ -9,6 +9,9 @@ blocks:
   estimateBlockTime: 2
   reorgPeriod: 1
 chainId: 84532
+deployer:
+  name: Michalis Kargakis
+  url: https://github.com/0xmichalis
```
paulbalaji, line 14:
```suggestion
  name: Toucan Protocol
  url: https://toucan.earth
```

chains/zora/metadata.yaml
```diff
@@ -9,6 +9,10 @@ blocks:
   estimateBlockTime: 2
   reorgPeriod: 1
 chainId: 7777777
+deployer:
+  email: merklylabs@proton.me
+  name: merklylabs
```
paulbalaji, line 14:
```suggestion
  name: Merkly Labs
```

chains/alfajores/metadata.yaml
```diff
@@ -13,6 +13,8 @@ blocks:
   estimateBlockTime: 5
   reorgPeriod: 0
 chainId: 44787
+deployer:
+  name: Abacus Works
```
jmrossy, line 17: What do you think of adding a `url: https://www.hyperlane.xyz` field to these too?

jmrossy (commented):
Looks good, one comment. I do wonder if we should make a non-required CI job to check if chains have deployer info. Otherwise chains for 3rd parties would likely forget to set this (the CLI doesn't prompt for this info)

chains/alfajores/metadata.yaml
paulbalaji: Yeah this was one thing I wasn't sure about - though we could I wasn't sure if we wanted to disambiguate between Hyperlane the protocol and AW the entity

jmrossy: It's a good question. When we get big enough, it's likely we'll want separate AW and Hyperlane web sites but for now maybe best to just include something that links the AW name back to official Hyperlane in some way

jmrossy (approved):

PR #103, opened 2024-07-14
Add Fhenix Testnet chain

0x471:
### Description
Add Fhenix Testnet chain.

Thanks to the Hyperlane Team—Yorke, Trevor, and all the hidden heroes—for your support! Greetings from ETHGlobal Brussels 😊

### Backward compatibility
Yes, these changes are backward compatible just by deleting the `fhenix` folder.

### Testing
I've tested using Fhenix as a destination chain. Not specifically with the deployed addresses in this PR.


chains/fhenix/addresses.yaml
```diff
@@ -0,0 +1,11 @@
+staticMerkleRootMultisigIsmFactory: "0x3fb9DC84b3c3eF6Ab0BeDC0835027cD2fF9800CE"
```
jmrossy, line 1: These will need to be ordered alphabetically. If you run yarn lint:fix it will take care of it for you

chains/fhenix/configs/core-config.yaml
```diff
@@ -0,0 +1,13 @@
+defaultHook:
```
jmrossy, line 1: We currently don't include configs in these registry

chains/fhenix/logo.jpg
jmrossy, line 1: Logos need to be SVGs. Is there an SVG version you can use (preferably a minified one with tools like [this](https://jakearchibald.github.io/svgomg/))

chains/fhenix/metadata.yaml
jmrossy, line 1: Would you mind adding a deployer field like this: https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/chains/alfajores/metadata.yaml#L16

jmrossy (commented):
Thanks for the contribution! Just a few things to fix before we can merge

jmrossy:
@0x471 Still a few little issues but I'll take this over for you and fix it up

PR #110, opened 2024-07-24
chore: update fraxtal transaction service

tkporter:
### Description

Taken from the Superchain interface as suggested by the Fraxtal team


Mo-Hussain:
Can you add a changeset please?

PR #119, opened 2024-07-30
fix(registry): update -config to -deployment in /deployments

brohamgoham:
### Description
Update Configuration File Naming Convention from *-config.yaml to *-deployment.yaml
This PR is in reference to [hyperlane-monorepo issue #4199 ](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4199
)

- [x] Changed the suffix from *-config.yaml to *-deployment.yaml when warp deployments are written to the registry's /deployments directory

- [x] Updated any associated references and documentation to ensure consistency with the new naming convention.
- [x] Update file names in /deployments

### Testing

- [x] Run registry unit tests


test/unit/registry.test.ts
brohamgoham, line 1: Registry Unit test fails with 1 error. 
Occurs when fetching github Registry
 `Fetches warp route configs for github registry`

jmrossy (commented):
Thanks @brohamgoham for the PR but this change would break compatibility with the existing Registry tooling deployed to NPM. Any users of the existing packages (including the Hyperlane core team), would have the warp route config fetching broken by the name change. Can I ask why you'd like to rename them? They're called `configs` because they contain the configs needed to create a `WarpCore`, the main class used to interact with warp routes

brohamgoham:
I was looking to take on the task in issue #4199. I apologize must have misunderstood the task, will delete now

jmrossy:
@brohamgoham Ah I see. I think maybe the specs in that ticket need some refinement first. Thanks for taking the initiative. We can find you something else ready for implementation if you'd like something.

brohamgoham:
@jmrossy yes please, I would love to contribute!

PR #120, opened 2024-07-31
fix: changeset ci check is no-op on release PRs

paulbalaji:
**problem**
the changeset release PRs get errors saying "Some packages have been changed but no changesets were found", because part of the process is moving the changeset contents to the readme and cleaning up all changesets. This is now a problem because the CI check expects there to be a changeset.

- fix: changeset ci check is no-op on release PRs


jmrossy (approved):
Thanks!

PR #124, opened 2024-08-01
fix: check changeset only on non-release PRs

paulbalaji:
- fix: check changeset only on non-release PRs
- will be skipped if branch is `changeset-release/main` or running in the merge queue


jmrossy (approved):
Much nicer :)

PR #132, opened 2024-08-06
Add Galadriel devnet

madschristensen99:
### Description
Add Galadriel Devnet

### Backward compatibility

Yes


chains/galadrieldevnet/logo.svg
jmrossy, line 1: This is not a valid SVG image file. Did you use a PNG or something like that? The logos must be vector graphics.

jmrossy (commented):
Mostly looks good but you need a changeset and the logo is not valid

chains/galadrieldevnet/logo.svg
madschristensen99: Yes sorry, I will adjust that file. I will make the changeset as well.

madschristensen99:
I have made the requested change to the logo, and I believe there is now a changeset. So we should be good to go with this merge.

jmrossy (approved):

PR #133, opened 2024-08-07
feat: improve haas metadata checks

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4231

- feat: improve haas metadata checks
- ensure AW mainnets have coingecko id
- try adding arb nitro marker in CI


scripts/check-for-coingecko-id.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
jmrossy, line 1: I realize this is already written so feel free to say no but I think it would be better for this check to be with the other unit tests. I think it would have been just a new lines of JS to check this in `test/unit/chains.test.ts`. WDYT?

paulbalaji: yup, you are absolutely right. originally had more in the script before toning down to just this check, but yeah it should be in the tests - will move it!

test/unit/chains.test.ts
```diff
@@ -13,6 +13,12 @@ describe('Chain metadata', () => {
     // it(`${chain} metadata contains deployer details`, () => {
     //   expect(metadata.deployer).not.to.be.undefined;
     // });
+
+    it(`${chain} metadata has gasCurrencyCoinGeckoId if deployer is Abacus Works it is a mainnet`, () => {
+      if (metadata.deployer?.name === "Abacus Works" && (metadata.isTestnet === undefined || metadata.isTestnet === false)) {
```
jmrossy, line 18:
```suggestion
      if (metadata.deployer?.name === "Abacus Works" && !metadata.isTestnet) {
```

jmrossy (approved):

PR #136, opened 2024-08-09
feat: script to check merkle tree hooks

paulbalaji:
- feat: script to check merkle tree hooks
- found super useful when trying to debug a recent chain deploy issue
- left the write-back option commented out for now


tkporter (approved):
Wondering if scripts like this make sense to have in the registry or somewhere else. Happy to merge if you think it'll be useful in the future

.changeset/spicy-ants-melt.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
jmrossy, line 2: Instead of this empty changeset we can use our admin perms to merge even without the changesets CI passing

jmrossy (approved):

paulbalaji:
> Wondering if scripts like this make sense to have in the registry or somewhere else. Happy to merge if you think it'll be useful in the future

Generally I think if something is generalisable to non-AW deployments I will try to have it here e.g. I plan to keep the sql-generation script in infra

PR #139, opened 2024-08-12
Add backup RPCs for some EVM chains

jmrossy:
### Description

Update RPCs for blast, cyber, fraxtal, inevm, manta, mode Soften explorer health test requirements

### Backward compatibility

Yes

### Related issue

https://github.com/hyperlane-xyz/issues/issues/1160#issuecomment-2284661052


paulbalaji (commented):
RPC additions are good but we can't remove worldchain as we have a live deployment there - it seems they've closed off public RPCs but allow private RPC usage

paulbalaji (approved):

PR #142, opened 2024-08-14
feat: Eclipse SOL token metadata

daniel-savu:


deployments/warp_routes/SOL/eclipse/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Eclipse SOL",
```
tkporter, line 2: I'd just call this "Solana" imo

tkporter (approved):

PR #148, opened 2024-08-19
chore: update zircuit explorer api

paulbalaji:
- added url based on https://docs.zircuit.com/dev-tools/verifying-contracts#automated-verification
- also added an api key to AW secrets
- drive-by sync rpc list with https://docs.zircuit.com/dev-tools/rpc-endpoints


tkporter (approved):
nice

PR #149, opened 2024-08-19
chore: rename fhenix to fhenixtestnet, update native token names

paulbalaji:
chore: rename fhenix to fhenixtestnet, update native token names


tkporter (approved):

jmrossy (approved):
LGTM

PR #155, opened 2024-08-20
fix: update Renzo warp artifacts

ltyu:
### Description
Adds the latest renzo artifacts. This is for the new deployment that fixes zircuit validator set

### Backward compatibility
Yes

### Testing
Tried on the UI, looks okay but still need Renzo to set the limits


ltyu:
same as https://github.com/hyperlane-xyz/hyperlane-registry/pull/153/files

PR #157, opened 2024-08-23
chore: lower the polygon confirmations

tkporter:
### Description

- Makes it 3, matching what we use internally in infra deployments https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/chains.ts#L23
- A response to https://discord.com/channels/935678348330434570/961710804011458621/1276321809872191572
- waiting 200 blocks for each transaction is unreasonable and results in timeouts in the CLI

### DIFF
--- a/.changeset/brave-gorillas-refuse.md
+++ b/.changeset/brave-gorillas-refuse.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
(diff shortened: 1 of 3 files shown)


paulbalaji:
Do we need to update the values for Amoy/zkevm too?

paulbalaji (approved):
pls add changese

PR #166, opened 2024-08-28
feat: Remove CoreChains

jmrossy:
### Description

Remove CoreChains enum and metadata maps since the only consumer (the explorer) will no longer need it soon. See https://github.com/hyperlane-xyz/hyperlane-explorer/pull/106

### Backward compatibility

No


paulbalaji (approved):
Once we go live with the latest batch of chain deployments, I'll be updating the registry used by the docs. I can remove references to CoreMainnets/CoreTestnets when I hit that

PR #168, opened 2024-08-29
feat: add an archive RPC node for Neutron

tkporter:
### Description

As Cosmos indexing requires an archive node, adding a public one


jmrossy (approved):
Nice

PR #171, opened 2024-08-30
chore: add `mainnet` suffix to solana and eclipse

daniel-savu:
### Description

This is a breaking change since `solana` and `eclipse` are renamed with the `mainnet` suffix, to allow for a new core deployment. See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4410


chains/eclipsemainnet/metadata.yaml
jmrossy, line 1: You'll need to rename the `name` field in this file too. Ditto for solana/metadata.yaml

daniel-savu: good spot, thanks

paulbalaji (approved):

PR #176, opened 2024-09-04
chore: remove unused testnets

paulbalaji:
chore: remove unused testnets


.changeset/green-cherries-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': minor
```
jmrossy, line 2: I think we established that removals should be major bumps

paulbalaji: nice catch, fat fingered it :D

jmrossy (approved):

PR #179, opened 2024-09-04
feat: Add proxyUrl generation

ltyu:
### Description
Adds logic to generate proxyUrl

### Backward compatibility
Yes - it is optional

### Testing
Manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "A collection of configs, artifacts, and schemas for Hyperlane",
   "version": "3.0.0",
   "dependencies": {
-    "yaml": "^2.4.5",
+    "yaml": "^2",
     "zod": "^3.21.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "lint-staged": "^15.2.2",
     "mocha": "^10.2.0",
     "prettier": "^2.8.8",
+    "sinon": "^18.0.0",
     "tsx": "^4.7.1",
     "typescript": "5.5.2",
     "typescript-eslint": "^7.7.0",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "mocha": "^10.2.0",
     "prettier": "^2.8.8",
     "tsx": "^4.7.1",
-    "typescript": "5.3.3",
+    "typescript": "5.5.2",
     "typescript-eslint": "^7.7.0",
     "zod-to-json-schema": "^3.22.5"
   },
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "A collection of configs, artifacts, and schemas for Hyperlane",
   "version": "3.0.0",
   "dependencies": {
-    "yaml": "^2",
+    "yaml": "2.4.5",
     "zod": "^3.21.2"
   },
   "devDependencies": {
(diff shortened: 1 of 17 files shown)


src/registry/GithubRegistry.ts
```diff
@@ -167,6 +170,11 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
     return `https://raw.githubusercontent.com/${this.repoOwner}/${this.repoName}/${this.branch}/${path}`;
   }
 
+  protected getApiUrl(): string {
+    const githubApiUrl = `https://api.github.com/repos/${this.repoOwner}/${this.repoName}/git/trees/${this.branch}?recursive=true`;
+    return this.proxyUrl ? `${this.proxyUrl}/${githubApiUrl}` : githubApiUrl;
```
jmrossy, line 175: So the url for a proxy would be like
`https://proxy.hyperlane.xyz/https://api.github.com/...` ? 
That looks odd to me. Does that work? I'd expect us to need to URL encode the path at the very least.

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "A collection of configs, artifacts, and schemas for Hyperlane",
   "version": "3.0.0",
   "dependencies": {
-    "yaml": "^2",
+    "yaml": "2.4.5",
```
ltyu, line 6: Upgraded to match monorepo so i can `yarn link`

src/registry/GithubRegistry.ts
ltyu: Updated to have without the entire path ie. https://proxy.hyperlane.xyz/repos/hyperlane-xyz/hyperlane-registry/git/trees/main?recursive=true

src/registry/GithubRegistry.ts
```diff
@@ -162,6 +177,27 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
     throw new Error('TODO: Implement');
   }
 
+  public async getApiUrl(): Promise<string> {
+    const { remaining, reset } = await this.getApiRateLimit();
```
ltyu, line 181: I know we discussed maybe not adding this, but i thought this would be useful anyways with minimum code.

Ontop of the extra quota that is per user IP, this provides a slightly better UX by allowing users to call 
`hyperlane-dev warp read --symbol eth --verbosity debug --registry  https://github.com/ltyu/hyperlane-registry`  without `--disableProxy`, given they have enough quota.

src/registry/GithubRegistry.ts
```diff
@@ -162,6 +177,27 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
...
+    }
+    return `${apiHost}/repos/${this.repoOwner}/${this.repoName}/git/trees/${this.branch}?recursive=true`
+  }
+
+  public async getApiRateLimit(): Promise<GithubRateResponse['resources']['core']> {
+    const response = await fetch(`${GITHUB_API_URL}/rate_limit`, {
+      headers: {
+        'X-GitHub-Api-Version': '2022-11-28'
+      }
+    })
```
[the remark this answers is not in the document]
ltyu: It's rate limited by a much higher secondary rate limit: 900 calls per minute https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits

jmrossy (approved):

PR #187, opened 2024-09-10
feat: deploy to everclear mainnet

paulbalaji:
feat: deploy to everclear mainnet


chains/everclear/logo.svg
jmrossy, line 1: This all-white logo will be invisible in most of the places we use it. It should be straightforward to change the path fill values to make it black. Lmk if you'd like help with that :)

paulbalaji: nice catch - updated!

PR #190, opened 2024-09-10
feat: deploy to oort mainnet

paulbalaji:
feat: deploy to oort mainnet


chains/metadata.yaml
```diff
@@ -2166,6 +2191,31 @@ osmosis:
   slip44: 118
   transactionOverrides:
     gasPrice: "0.025"
+piccadilly:
```
tkporter, line 2194: intentional? I don't see it in the `chains` dir

tkporter (approved):

chains/metadata.yaml
paulbalaji: yup it's here https://github.com/hyperlane-xyz/hyperlane-registry/tree/pb/oort/chains/piccadilly

paulbalaji: from https://github.com/hyperlane-xyz/hyperlane-registry/pull/156

PR #201, opened 2024-09-16
chore: add test to ensure nitro chains define index.from

paulbalaji:
chore: add test to ensure nitro chains define index.from


package.json
```diff
@@ -10,6 +10,7 @@
     "@changesets/cli": "^2.26.2",
     "@eslint/js": "^9.1.1",
     "@hyperlane-xyz/sdk": "4.1.0",
+    "@hyperlane-xyz/utils": "4.1.0",
```
jmrossy, line 13: The downside of adding this explicitly is that devs must be careful to increment this and the SDK together - whereas if it's implicitly included via the SDK's transitive dep that won't be an issue.

The upside is that it will improve code editor support.

paulbalaji: ah true, will update and leave it as a transitive dep

jmrossy (approved):

PR #204, opened 2024-09-18
feat: turbo eth synthetic token

daniel-savu:
Reuses the eth logo until we get a custom tETH one


deployments/warp_routes/tETH/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Turbo ETH",
+  "symbol": "tETH",
+  "description": "Warp Route tETH",
+  "image": "",
```
paulbalaji, line 5: no image?

tkporter, line 5: does this need to point to the logo or no?

tkporter (approved):

deployments/warp_routes/tETH/metadata.json
daniel-savu: couldn't find any logo, just asked them about it. The warp UI can be updated when one becomes available luckily

daniel-savu: this is a custom format I used for synthetic metadata based on an example somewhere, all fields are optional

paulbalaji (approved):

PR #208, opened 2024-09-18
feat: WIF token metadata

daniel-savu:
- the WIF logo is not available as svg, used jpg
- sets the Turbo ETH logo as a drive-by


paulbalaji (approved):
hat stays on

PR #220, opened 2024-09-19
fix(warp): Update solana eclipse USDC warp route addresses

Mo-Hussain:
### Description
- update USDC solana, ethereum, eclipse warp config


paulbalaji (dismissed):

daniel-savu (approved):
fire

PR #226, opened 2024-09-19
changed `displayName` to Aleph Zero EVM

avious00:
requested by Aleph Zero contributors to avoid confusion with original Aleph Zero L1


chains/alephzero/metadata.yaml
```diff
@@ -11,7 +11,7 @@ chainId: 41455
 deployer:
   name: Abacus Works
   url: https://www.hyperlane.xyz
-displayName: Aleph Zero
+displayName: Aleph Zero EVM
```
jmrossy, line 14: I recommend also setting a `displayNameShort`

jmrossy (approved):

PR #233, opened 2024-09-30
chore: add deployer name for AW testnets

paulbalaji:
chore: add deployer name for AW testnets

did a pass on all chains, seems only recent testnets were affected

### DIFF
--- a/.changeset/two-boxes-compete.md
+++ b/.changeset/two-boxes-compete.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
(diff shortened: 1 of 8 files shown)


tkporter (approved):
should this be required by CI?

paulbalaji:
> should this be required by CI?

we disabled it originally because throwing up metadata-only PRs without addresses is valid, and we wouldn't want to force a deployer to be defined for that

could probs update the test to only check if there's a mailbox set

paulbalaji:
merging to get a patch release out to update docs

PR #236, opened 2024-09-30
Update tangletestnet addresses and add WETH warp route

devpavan04:
### Description

- Updates `tangletestnet` addresses and metadata.
- Adds a new warp route - `WETH` from `holeksy` to `tangletestnet`.


devpavan04:
@paulbalaji Below is the core config for new tangletestnet deployments:

CleanShot 2024-09-30 at 11 18 11

I have also updated the warp route to include the correct relayer address:

CleanShot 2024-09-30 at 11 20 49

Since we're using trustedRelayerISM here, I am trying to run just one relayer.

CleanShot 2024-09-30 at 11 25 11

I have few questions:

1. I want to transfer WETH from holesky to tangletestnet. Do I need to include both the chains when generating the agent config? or just the destination chain? The docs specify `--chains chain1` which is confusing.

CleanShot 2024-09-30 at 11 27 26

2 .When trying to generate agent config for `--chains tangletestnet` I see these warnings:

CleanShot 2024-09-30 at 11 32 11

Could you please tell me how to get this working and if I am missing something here. I am following [Deploy Hyperlane with Local Agents ](https://docs.hyperlane.xyz/docs/guides/deploy-hyperlane-local-agents) from the docs.

paulbalaji:
overall LGTM, please run `yarn changeset add` to add a `minor` changeset

paulbalaji:
> I want to transfer WETH from holesky to tangletestnet. Do I need to include both the chains when generating the agent config? or just the destination chain? 

The relayer needs to have metadata on the chains that it should watch/deliver to, so if you want the relayer to go between chain1/chain2 it should have both in the config

> When trying to generate agent config for --chains tangletestnet I see these warnings:

When you deployed the core contracts there should be a `merkleTreeHook` that was also deployed. There's currently a bug where this doesn't get correctly to the artifacts in some cases. If you add that `merkleTreeHook` address, it should be able to generate the agent config

I believe this is the one you deployed https://testnet-explorer.tangle.tools/address/0x1d7A5b6e18c16103c365e975B258Cb5039971E9b

devpavan04:
> I believe this is the one you deployed https://testnet-explorer.tangle.tools/address/0x1d7A5b6e18c16103c365e975B258Cb5039971E9b

Yes that seems correct.

@paulbalaji Thanks for the response here! One more question.

For the transfer to work, does the tangletestnet related config should be in core config of holesky?

`hyperlane core read --chain holesky` does not include tangletestnet in it's defaultHook, defaultISM domains.

devpavan04:
> Yup that's right, if you want to use the existing mailbox we have there we'd have to enroll tangletestnet into the default chains there.
> 
> You can also deploy your own mailbox/contracts on holesky, and the trusted relayer could relay between your two mailboxes

Okay. So deploy to mailbox and other contracts on holesky, it's the same process right.

`hyperlane registry init` and `hyperlane core init` and `hyperlane core deploy`.

Also when deploying on holesky, can I just do `hyperlane core init` to keep core config to default or `hyperlane core init --advanced` ?

I believe after deploying the contracts on holesky, when I generate agent config:

`hyperlane registry agent-config --chains holesky,tangletest`

I need to make sure the core contracts on holesky points to the addresses I just deployed.

Also, `hyperlane core deploy` does not deploy `interchainGasPaymaster` contract. So this will not work right? How will the sender on holesky send the gas fee to paymaster so that relayer can use it on tangletestnet? Is this a bug as well - not able to deploy `interchainGasPaymaster` with `hyperlane core deploy`?

Also the docs mention, there needs to be 1 paymaster contracts for 1 relayer. Not sure how to do this.

chains/tangletestnet/metadata.yaml
```diff
@@ -1,23 +1,21 @@
...
 blocks:
-  confirmations: 1
+  confirmations: 4
   estimateBlockTime: 6
-  reorgPeriod: 0
+  reorgPeriod: 4
 chainId: 3799
-deployer:
-  name: Abacus Works
-  url: https://www.hyperlane.xyz
```
jmrossy, line 13: Can you please set this to a new name and URL

paulbalaji:
It should be the same process yes

> I need to make sure the core contracts on holesky points to the addresses I just deployed.

yup, if you have your local registry you can use `--overrides <pathtoregistry>` with the cli to override the default addresses 

> Also, `hyperlane core deploy` does not deploy `interchainGasPaymaster` contract. So this will not work right? How will the sender on holesky send the gas fee to paymaster so that relayer can use it on tangletestnet? Is this a bug as well - not able to deploy `interchainGasPaymaster` with `hyperlane core deploy`?
> Also the docs mention, there needs to be 1 paymaster contracts for 1 relayer. Not sure how to do this.

Yeah I'm looking into the IGP cli situation at the moment. To unblock you running in the interim, worth trying to set `interchainGasPaymaster` to the zero address

PR #270, opened 2024-10-10
Update Cheesechain Routes

bivsk:
### Description

Update CheeseChain USDC route


jmrossy (changes requested):
A few problems here:
1) if the route is cheesechain <-> arbitrum, it should not be changing this existing cheesechain <-> ethereum file
2) The PR title is inaccurate  
3) You're missing a changeset (run `yarn changeset`)

bivsk:
Regarding 1, we do not want to support the old cheesechain <-> ethereum route any longer. Can I just rename this current file?

rshtirmer:
> A few problems here:
> 
> 1. if the route is cheesechain <-> arbitrum, it should not be changing this existing cheesechain <-> ethereum file
> 2. The PR title is inaccurate
> 3. You're missing a changeset (run `yarn changeset`)

Hey J,

I published the original route that @bivsk is overwriting - that route being overwritten is deprecated, and does not point to what is now the canonical USDC on cheesechain (which is created by bridging USDC from arb -> cheesechain via this new route)  - @bivsk will add a changeset, but otherwise I think this should be ok? Please let me know if theres anything else we can do to support. 

Best,
Ryan Shtirmer

bivsk:
@jmrossy I've made a few changes to hopefully fix the above issues, while also sticking to our plan to replace rather than add a route.  Hopefully this works, but let me know if not.  Thanks a lot, and have a good rest of your weekend

.changeset/chilled-yaks-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
```
jmrossy, line 2:
```suggestion
'@hyperlane-xyz/registry': major
```

jmrossy: Whenever something is removed, that's a breaking change and the version increment will need to be `major`

deployments/warp_routes/USDC/cheesechain-arbitrum-config.yaml
jmrossy, line 1: Ditto above for filename

deployments/warp_routes/USDC/cheesechain-arbitrum-addresses.yaml
jmrossy, line 1: The filename is now incorrect. The chains should be alphabetically ordered (arbitrum-cheesechain).

jmrossy (commented):
Almost there!

bivsk:
OK pretty sure it's all good to go now 😅

Thanks again!

jmrossy:
@bivsk Still one more problem. run `yarn lint:fix`

Screenshot 2024-10-20 at 12 50 18 PM

jmrossy (approved):

PR #274, opened 2024-10-10
feat: add arcadiatestnet, sonictestnet, unichaintestnet

paulbalaji:
### Description

Add deployments for testnets: arcadiatestnet, sonictestnet, unichaintestnet

### Backward compatibility

yes

### Testing

manual


yorhodes (approved):
fwiw this is why I think the ISM/hook addresses should not be in artifacts (because they change)

paulbalaji:
> fwiw this is why I think the ISM/hook addresses should not be in artifacts (because they change)

yeah I'm coming to agreement! Had a few support requests where people get confused why the onchain defaultISM might differ from the one in registry

looking to tidy up in future 🤞

PR #287, opened 2024-10-14
feat: Add Sei to Renzo

ltyu:
### Description
Add Sei to Renzo

### Backward compatibility
Yes

### Testing
- Tested using hyperlane cli to send a tx


yorhodes (approved):
I like how clean this diff is :)

PR #289, opened 2024-10-15
feat: Add configuration for currencies supported on particular chain

ameten:
### Description

Add configuration for currencies supported on particular chain. The configuration allows to specify mapping from the identifier of crypto currency which is used in transaction object to UUID which uniquely identify this crypto currency for Hyperlane agent.

### Related issues

- Contributes to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4641

### Backward compatibility

Yes

### Testing

Generated agent config (`mainnet-config.json`) and started Scraper with it just to check that new field does not break it.

### DIFF
--- a/chains/metadata.yaml
+++ b/chains/metadata.yaml
@@ -2435,6 +2435,17 @@ neutron:
     - http: https://rpc-arch.mainnet.neutron.tm.p2p.org
     - http: https://rpc-kralum.neutron-1.neutron.org
   slip44: 118
+  supportedCurrencies:
+    - description: It is not TIA on Celestia chain. It is TIA bridged to Neutron chain.
+      identifier: ibc/773B4D0A3CD667B2275D5A4A7A2F0909C0BA0F4059C0B9181E680DDF4965DCC7
+      symbol: TIA
+      uuid: 0b03525c-ec50-4176-9318-b70636e0c054
+    - identifier: ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9
+      symbol: ATOM
+      uuid: 5b606c8a-795b-4d82-ad8e-4181943a2ef9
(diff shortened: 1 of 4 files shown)


chains/metadata.yaml
```diff
@@ -2435,6 +2435,17 @@ neutron:
     - http: https://rpc-arch.mainnet.neutron.tm.p2p.org
     - http: https://rpc-kralum.neutron-1.neutron.org
   slip44: 118
+  supportedCurrencies:
```
daniel-savu, line 2438: I'm torn on the naming, let's discuss in the notion doc

chains/neutron/metadata.yaml
```diff
@@ -34,3 +34,14 @@ rpcUrls:
 slip44: 118
 transactionOverrides:
   gasPrice: "0.0075"
+supportedCurrencies:
+  - symbol: TIA
+    description: "It is not TIA on Celestia chain. It is TIA bridged to Neutron chain."
+    identifier: ibc/773B4D0A3CD667B2275D5A4A7A2F0909C0BA0F4059C0B9181E680DDF4965DCC7
```
jmrossy, line 40: `addressOrDenom` may be more consistent with other field names

chains/neutron/metadata.yaml
```diff
@@ -34,3 +34,14 @@ rpcUrls:
...
+    uuid: 0b03525c-ec50-4176-9318-b70636e0c054
```
jmrossy, line 41: Is this just randomly generated when you put together this config?

chains/neutron/metadata.yaml
```diff
@@ -34,3 +34,14 @@ rpcUrls:
 slip44: 118
 transactionOverrides:
   gasPrice: "0.0075"
+supportedCurrencies:
```
jmrossy, line 37: There's already a `nativeToken` field. I wonder if this should be folded into that? Or at least named `supportedTokens` or something or consistency?

jmrossy: Or `supportedGasTokens`

PR #300, opened 2024-10-17
feat(amphrETH): add warp route

daniel-savu:
### Description

Adds ethereum <> zircuit amphrETH warp route artifacts


deployments/warp_routes/AMPHRETH/ethereum-zircuit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# yaml-language-server: $schema=../schema.json
+tokens:
+  - addressOrDenom: "0xA4e45cB601d43818518FB6023b8070d2564b935f"
```
jmrossy, line 3: Not mandatory but you probably want a `logoURI` field in these token entries if they're gonna be used in a UI

daniel-savu: we don't have one for now, will just default to whatever the UI uses

PR #305, opened 2024-10-18
fix: update inevm fallbackRoutingHook address

paulbalaji:
### Description

fix: update inevm fallbackRoutingHook address

drive-by: update the hook address checking script
- only check AW chains
- only log when there's a geniune problem

### Backward compatibility

yes

### Testing

manual


scripts/check-mailbox-hooks.sh
```diff
@@ -22,9 +22,15 @@ for chain_dir in chains/*/; do
     metadata_file="$chain_dir/metadata.yaml"
     addresses_file="$chain_dir/addresses.yaml"
 
+    # Check if the chain is an Ethereum protocol type
+    if ! yq e '.deployer.name == "Abacus Works"' "$metadata_file" | grep -q true; then
+        # echo "$chain_name: Not Abacus Works, skipping."
```
tkporter, line 27: nit: do we want this? ditto for the other 3 commented out echos below

tkporter (approved):

scripts/check-mailbox-hooks.sh
paulbalaji: ah rip i thought i added them back again, will update

PR #315, opened 2024-10-22
minimal fix: Resolve RPC error in chain configuration (worldchain)

Fazleiw:
### Description

This fix addresses an RPC error that was causing issues during deployments. The issue was related to a misconfiguration in the chain's RPC endpoints, and it has been resolved by overriding the correct RPC details. All configurations have been tested and deployed using the Hyperlane tooling.

- **Summary of change**: Fixes the bug due to an RPC misconfiguration.

### Backward compatibility

yess

### Testing

The fix has been tested with existing metadata configs and deployment addresses using Hyperlane CLI. All tests passed successfully.


chains/worldchain/metadata.yaml
```diff
@@ -22,5 +22,6 @@ nativeToken:
   symbol: ETH
 protocol: ethereum
 rpcUrls:
+  - http: https://worldchain-mainnet.g.alchemy.com/public
   - http: https://raas-backend.alchemy.com/rpc/worldchain-mainnet/rollup
```
paulbalaji, line 26:
```suggestion
  - http: https://worldchain-mainnet.g.alchemy.com/public
```

paulbalaji (commented):
please also add a `patch` changeset

PR #321, opened 2024-10-23
Update logo.svg

LyDe17TC:
I changed the clique logo to the one that appears on their website


paulbalaji:
these are different cliques

PR #322, opened 2024-10-23
Update logo.svg

LyDe17TC:
I changed the clique logo to the one that appears on their site


paulbalaji:
these are different clique projects.

in the registry we have: https://www.myclique.io/
the one you are changing to is https://www.clique.tech/

LyDe17TC:
ok sorry

PR #320, opened 2024-10-22
Update logo.svg

LyDe17TC:
I've changed the logo to the one that seems to be used most by the project

PR #323, opened 2024-10-23
adding the apechain

LyDe17TC:
add apechain chain

some information is missing due to the newness of the channel, including the addresses file

I used the cli to create the metadata file


paulbalaji:
we'll be doing our own deployment to apechain in the future, so will close this PR for now - i see the citrea logo update is already in another PR waiting on action: https://github.com/hyperlane-xyz/hyperlane-registry/pull/320

LyDe17TC:
Ok no problems

PR #333, opened 2024-10-25
feat: initial metadata for oct-25 batch

paulbalaji:
### Description

feat: initial metadata for oct-25 batch

- add metadata for new chains
- remove any existing deployment addresses for chains that re going to join the core set
- update tests for sanity checking AW-chain metadata
- remove deployer field from metadata-only chains that are not in immediate AW plans

### Backward compatibility

yes

### Testing

locally


chains/apechain/logo.svg
jmrossy, line 1: There may be horizontal dead-space around the SVG here
https://svgcrop.com/

.changeset/nasty-emus-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': minor
```
jmrossy, line 2: Since something was removed (fantom addresses), this should be major.

chains/gravity/metadata.yaml
```diff
@@ -0,0 +1,28 @@
...
+    url: https://explorer.gravity.xyz
+blocks:
+  confirmations: 1
+  estimateBlockTime: 1
+  reorgPeriod: 5
+chainId: 1625
+deployer:
+  name: Abacus Works
+  url: https://www.hyperlane.xyz
+displayName: Gravity Alpha Mainnet
```
jmrossy, line 15:
```suggestion
displayName: Gravity Alpha
```

chains/gravity/metadata.yaml
```diff
@@ -0,0 +1,28 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://explorer.gravity.xyz/api
+    family: blockscout
+    name: Gravity Alpha Mainnet Explorer
```
jmrossy, line 5:
```suggestion
    name: Gravity Alpha Explorer
```

chains/gravity/metadata.yaml
jmrossy: Or set a displayNameShort field

test/unit/chains.test.ts
```diff
@@ -4,6 +4,8 @@ import { ChainAddressesSchema } from '../../src/types.js';
 
 import { expect } from 'chai';
 
+const ABACUS_WORKS_DEPLOYER_NAME = "Abacus Works";
```
jmrossy, line 7: Do you think we should put this in `src` and export it properly? I was on the fence about it (for neutrality) but it seems like we'll use it in a few places

jmrossy (commented):
mostly looks good!

chains/apechain/logo.svg
paulbalaji: ah thanks i did stick it into the crop tool but i didn't copy the output before going to the minifier https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/892a1d82826e2ca6b25e791259065d9a6dd17952/typescript/sdk/src/deploy/verify/ContractVerifier.ts#L142-L159

.changeset/nasty-emus-vanish.md
paulbalaji: they're being replaced with our own addresses in the upcoming PR which addresses for these new chains, would that still mean this is a minor?

chains/gravity/metadata.yaml
paulbalaji: I'll set `displayNameShort` to `Gravity`

test/unit/chains.test.ts
paulbalaji: yeah I think it makes sense, and probably include an `isAbacusWorksChain` helper function to clean things up

.changeset/nasty-emus-vanish.md
jmrossy: If we don't release in between then it's probably. Tbh it's probably fine either way, just not _strictly_ ideal

jmrossy (approved):

PR #346, opened 2024-10-30
Update metadata.yaml

JunayedMh:


JaunePomme:
Not sure if it will pass, contribution is being noticed, this is just a detail to the .yml, but meh, who knows

PR #349, opened 2024-10-30
feat: eclipse batch 2 token metadatas

daniel-savu:
### Description

ORCA, WBTC, USDT


deployments/warp_routes/ORCA/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Orca",
+  "symbol": "ORCA",
+  "description": "Warp Route ORCA",
+  "image": "https://raw.githubusercontent.com/hyperlane-xyz/hyperlane-registry/refs/heads/main/deployments/warp_routes/ORCA/logo.svg",
```
tkporter, line 5: should this be pinned to a certain commit? same for the other 2

as it stands now
```
$ curl https://raw.githubusercontent.com/hyperlane-xyz/hyperlane-registry/refs/heads/main/deployments/warp_routes/ORCA/logo.svg
404: Not Found
```

For the other ones despite them being in main already a permalink seems safest just in case we shuffle around the registry

daniel-savu: discussed that options are
- 2 step merge to get proper permalinks
- use permalinks with hashes of commits from a PR. This depends on GitHub's garbage collection policy, which seems to differ from that of the `git` spec as seemingly removed branches don't have the commits removed - making permalinks still retrievable. It's unclear if those commits ever get garbage collected, as they normally would
- keep referencing non-permalinked paths
- new repo just for logos to get permalinks more easily

for now we're going ahead with the approach used so far of referencing non-permalinked paths

tkporter (approved):

PR #352, opened 2024-10-31
feat: migrate metadata from sept26 chain deployment

paulbalaji:
### Description

only merge this AFTER https://github.com/hyperlane-xyz/hyperlane-explorer/pull/136 is merged

Migrate metadata for sept26 chain deployment batch
- `immutablezkevm` -> `immutablezkevmmainnet`
- `rari` -> `rarichain`
- `rootstock` -> `rootstockmainnet`
- `alephzeroevm` -> `alephzeroevmmainnet`
- `chiliz` -> `chilizmainnet`
- `lumia` -> `lumiaprism`
- `superposition` -> `superpositionmainnet`
- `flow` -> `flowmainnet`
	- unfortunately `flowevm` was burned during deployment
- `metall2` -> `metal`
- `polynomial` -> `polynomialfi`

new domainids all have `10000` prefixed

also updates the gecko ID for lumia

### Backward compatibility

not really, all the deprecated chains have a new name now

### Testing

tbd, manual


chains/azeroevm/metadata.yaml
```diff
@@ -13,11 +13,11 @@ deployer:
   url: https://www.hyperlane.xyz
 displayName: Aleph Zero EVM
 displayNameShort: Aleph Zero EVM
-domainId: 41455
+domainId: 4294941455
```
tkporter, line 16: wdyt of like `1000041455` etc? I think it'd be easier to eyeball them, which I think is important

tkporter (commented):
These are my only suggestions for name changes - I think some of them rn aren't super recognizable, and I'm more comfortable going longer than having chains where it's not obvious which they are:

immutablezkevm -> immutzkevm or immutablezkmainnet
rootstock -> rootstockmainnet
alephzeroevm -> alephzroevm or alephzevm
chiliz -> chilizmainnet
superposition -> superpositionmainnet

chains/azeroevm/metadata.yaml
paulbalaji: works for me

paulbalaji:
if we're ok with longer names then i think we can just make them all `chain` -> `chainmainnet`

chains/lumia/metadata.yaml
daniel-savu, line 1: should this file's parent dir be renamed too?

daniel-savu (approved):

chains/lumia/metadata.yaml
paulbalaji: not yet - because we'll be keeping lumia+lumiaprism live together for a little bit

PR #354, opened 2024-11-01
fix: unit test flake

daniel-savu:
### Description

increases timeout of unit test that takes longer now that we have more warp routes


test/unit/registry.test.ts
```diff
@@ -96,7 +96,7 @@ describe('Registry utilities', () => {
       expect(firstRoute!.tokens.length).to.be.greaterThan(0);
       const noRoutes = await registry.getWarpRoutes({ symbol: 'NOTFOUND' });
       expect(Object.keys(noRoutes).length).to.eql(0);
-    });
+    }).timeout(10_000);
```
tkporter, line 99: is this seconds? or ms

tkporter (approved):

test/unit/registry.test.ts
daniel-savu: milliseconds, you can check the other configured timeouts in this file. The default is `2_000` and that gets exceeded pretty frequently for this test now

PR #355, opened 2024-11-01
fix: eclipse warp route fixes

daniel-savu:
### Description
a path was wrong, some logos were missing, a file was named incorrectly


deployments/warp_routes/USDT/eclipsemainnet-ethereum-solana-config.yaml
```diff
@@ -6,6 +6,7 @@ tokens:
     connections:
```
tkporter, line 6: name should be solanamainnet, not solana

tkporter (approved):

PR #357, opened 2024-11-04
feat: ensure domainIds are required

paulbalaji:
### Description

add test to ensure chains have domainId set

### Backward compatibility

effectively a metadata change without a schema update so shouldn't break back compat as the schema is unchanged

### Testing

ci passes

### DIFF
--- a/.changeset/modern-ducks-shake.md
+++ b/.changeset/modern-ducks-shake.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': major
+---
+
+Ensure domainIds are required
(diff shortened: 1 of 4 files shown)


tkporter (approved):
Is this not caught by the zod schema? if not should it?

paulbalaji:
> Is this not caught by the zod schema? if not should it?

it will be after https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798/files#diff-27e454a48553edc89911986f62741edf834c8f12a00db4d549d3a1c150ba31edL190

PR #358, opened 2024-11-04
feat: improve warp route unit tests

paulbalaji:
### Description

- add test for warp route ids
- ensure chain names are in alphabetical order
- ensure each chain name is in the registry
- ensure each token logoURI points to a valid path in registry

### Testing

tests pass


.changeset/olive-spoons-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
```
paulbalaji, line 2: major because of warp name changes?

test/unit/warp-routes.test.ts
```diff
@@ -17,5 +21,31 @@ describe('Warp Route Configs', () => {
...
+    it(`Route ${id} has valid chain names`, () => {
+      const { chainNames } = parseWarpRouteConfigId(id);
+
+      // Verify each chain exists in registry
+      for (const chain of chainNames) {
+        expect(localRegistry.getChainMetadata(chain), `Chain ${chain} not found in registry`).to.not.be.null;
+      }
+    });
+
+    it(`Route ${id} has chain names in alphabetical order`, () => {
```
tkporter, line 34: wonder if there should also be a test of "deriving the warp ID from the config is equal to `id`", which iiuc would also have covered some of these cases

.changeset/olive-spoons-return.md
tkporter: 🤷‍♂️ up to you, I'm personally happy with a minor but I'm the resident semver ignorer

tkporter (approved):

PR #370, opened 2024-11-07
feat: add coinGeckoId to tokens in warp configs

tkporter:
### Description

This adds a `coinGeckoId` to get an approximate USD value of the token. Some tokens are not yet listed on Coingecko, so an appropriate substitute is chosen.

The motivation for this is https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4835 -- we are removing the special "deployment yaml" that was used by the warp monitor service in favor of WarpCore.

~~The corresponding zod change is in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4835 here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4835/files#diff-5d9f643079328fbbe587e59bc31d809c46b98037c045d2df3931b22acf87691fR50 and hasn't yet been released. I'm happy to get that zod change in and do an SDK release first if preferred, though as extra fields are permitted by our types I am cheekliy trying to save myself the hassle~~

This needs to be merged first https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4842, then this PR, then https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4835

### Backward compatibility

Backward compatible (our schemas allow extra params, see above, happy to do the sdk release though)


tkporter:
Seems `yarn build` is forcing my hand to publish the SDK first 😭

paulbalaji (approved):

PR #376, opened 2024-11-12
Add reactive testnet

Zorato:
### Description

Add Kopli Reactive (testnet) chain. EVM compatible.


chains/reactivekopli/metadata.yaml
```diff
@@ -0,0 +1,22 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://kopli-api.reactscan.net/
```
JaunePomme, line 3: take the last "/" off I think

chains/reactivekopli/metadata.yaml
```diff
@@ -0,0 +1,22 @@
...
+    family: blockscout
+    name: reactscan
+    url: https://kopli.reactscan.net/
```
JaunePomme, line 6: Same, take the last "/" off I think

JaunePomme:
Modify your metadata.yml a bit
Could add the deployer name in the metadata.yml
Run prettier
Minimify the svg, here is the link : https://jakearchibald.github.io/svgomg/
And check other PR cuz i'm not sure you need the height and width in it.
Then I think it's good !

daniil-parsiq:
Thanks for the review, I fixed everything according to your comments.

chains/reactivekopli/metadata.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://kopli-api.reactscan.net
+    family: blockscout
```
paulbalaji, line 4:
```suggestion
    family: other
```

chains/reactivekopli/metadata.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://kopli-api.reactscan.net
```
paulbalaji, line 3: are there any docs r.e. the reactscan explorer API

daniil-parsiq: well... there really isn't. but we are ready to add necessary methods on a short notice, question is - where we can find some sort of definition for required methods

paulbalaji: typically most explorers will have some endpoint that is compatible with the etherscan api spec https://docs.etherscan.io/api-endpoints/contracts#verify-solidity-source-code

e.g. blockscout docs point here https://docs.blockscout.com/devs/apis/rpc

paulbalaji (approved):

PR #379, opened 2024-11-13
Add cosmoshub chain definition

jmrossy:
### Description

Add chain metadata for CosmosHub.
This is used in the Warp UI as the 'default' cosmos chain if no others are needed for the given routes. We need a default or the Cosmos libs throw errors.

### Backward compatibility

Yes


chains/cosmoshub/metadata.yaml
```diff
@@ -0,0 +1,34 @@
+# yaml-language-server: $schema=../schema.json
+bech32Prefix: cosmos
+blockExplorers:
+  - apiUrl: https://www.mintscan.io/cosmos
```
xeno097, line 4: Is this correct here? I see they have these docs for the API https://docs.cosmostation.io/apis

jmrossy: It doesn't actually matter b.c. we don't use apiUrls for explorer families other than Etherscan atm. But you're correct that there's now a real api so I've updated all the minscan block explorers :)

xeno097 (approved):

PR #396, opened 2024-11-25
feat: deploy to B² Network

paulbalaji:
### Description

feat: deploy to B² Network


xeno097 (approved):
just curious why so many ism addresses changes?

paulbalaji:
> just curious why so many ism addresses changes?

mentioned in description: drive-by update to default ISMs for boba, duckchain, superseed, unichain, vana as they are also in the same batch as bsquared

so that we can test message sending between all 5 of these chains

PR #404, opened 2024-11-28
feat: deploy to swell & lumiaprism

paulbalaji:
### Description

- feat: deploy to swell & lumiaprism
- drive-by: updates default ISMs for the other chains in the batch as well

### Testing

manual


chains/addresses.yaml
```diff
@@ -944,6 +944,21 @@ cyber:
   testRecipient: "0x2c61Cda929e4e2174cb10cd8e2724A9ceaD62E67"
   timelockController: "0x0000000000000000000000000000000000000000"
   validatorAnnounce: "0x062200d92df6bb7ba89ce4d6800110450f94784e"
+deepbrainchaintestnet:
```
Mo-Hussain, line 947: should we expect this change?

Mo-Hussain (approved):

chains/addresses.yaml
paulbalaji: yup - because this file wasn't autoupdated with the community PR

PR #405, opened 2024-11-29
feat: add mitosis testnet

JunayedMh:
### Description
add mitosis testnet
docs : https://docs.mitosis.org/docs/learn/game-of-mito/setup
minified by: https://jakearchibald.github.io/svgomg/

### Testing
yes


chains/mitosistestnet/metadata.yaml
```diff
@@ -0,0 +1,21 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://testnet.mitosiscan.xyz/api
+    family: routescan
```
JaunePomme, line 4: Not sure about that family name. I think it would be more blockscout or something else

JaunePomme:
Overall looks good to me, besides the family name, and dunno if you minimified the svg (https://jakearchibald.github.io/svgomg/).

chains/mitosistestnet/metadata.yaml
JunayedMh: Yeah it is routerscan
Chk on the explorer

k4ctuss: you should make family name to "other" cause yours isn't propose by the CLI

paulbalaji (approved):

PR #416, opened 2024-12-10
fix: use correct apxETH logo

tkporter:
### Description

A copy&paste mishap


chains/inclusivelayertestnet/metadata.yaml
```diff
@@ -23,3 +23,4 @@ nativeToken:
 protocol: ethereum
 rpcUrls:
   - http: http://192.241.245.190:18449
+technicalStack: arbitrumnitro
```
paulbalaji, line 26: this was probably auto-added, but think you'll need to add an index.from field too 

http://192.241.245.190/tx/0x7376b707831c3b91b325768bbff046eba81062a72d42190e1f8094d652943d9b

think `4306`

paulbalaji (approved):

PR #430, opened 2024-12-15
feat: add VANA warp route between Ethereum and Vana

tkporter:
### Description

- Deployed by Protofire


tkporter:
need to add coinGeckoId later

PR #445, opened 2024-12-18
ci: add workflow to check and optimize svgs

Xaroz:
### Description

- Add workflow to trigger script to validate and optimize SVGs
- Add script to validate SVG and then optimize
- Added [SVGO library ](https://www.npmjs.com/package/svgo), which helps optimize SVGs

PD: Many svgs were optimized, but it shouldn't trigger again since no changes will be needed

(Tested with Max 20KBs but it should be 100 KBs)
Screenshot 2024-12-23 at 7 48 03 AM

### Backward compatibility
Yes

### Testing

CLI testing and testing in warp ui


scripts/optimize-svg.js
```diff
@@ -0,0 +1,44 @@
...
+
+  if (!fileName.endsWith('logo.svg')) {
+    console.error(`Error: File does not end with 'logo.svg' -> ${filePath}`);
+    process.exit(1); // Exit immediately if criteria is not met
+  }
+
+  if (stats.size > MAX_FILE_SIZE) {
+    console.error(`Error: File size exceeds 20KBs -> ${filePath}`);
+    process.exit(1); // Exit immediately if criteria is not met
+  }
```
paulbalaji, line 19: print the actual file size as well

Xaroz: done

package.json
```diff
@@ -22,6 +22,7 @@
     "mocha": "^10.2.0",
     "prettier": "^2.8.8",
     "sinon": "^18.0.0",
+    "svgo": "^3.3.2",
```
paulbalaji, line 25: i had no idea the website was just a prettified UI around this

scripts/optimize-svg.js
```diff
@@ -0,0 +1,108 @@
...
+
+  const fileContent = fs.readFileSync(filePath, 'utf8');
+  if (RASTER_IMAGE_REGEX.test(fileContent)) {
+    console.error(
+      `Error: File contains an <image> tag, likely embedding a raster image -> ${filePath}`,
+    );
+    process.exit(1);
+  }
+}
```
paulbalaji, line 33: I think it would be better if we catch all the invalid SVGs and print one output at the end, instead of failing fast - wdyt?

package.json
Xaroz: Yeah! Also this gave me the idea of locking the version, so that files are not modified often because of changes between version

scripts/optimize-svg.js
Xaroz: Done, instead of exiting, it just prints the error and then exit at the end if there an error

scripts/optimize-svg.js
```diff
@@ -0,0 +1,112 @@
+import fs from 'fs';
+import path from 'path';
+import { optimize } from 'svgo';
+
+const directories = ['./chains', './deployments'];
+const MAX_FILE_SIZE = 100 * 1024; // 100KBs
+const RASTER_IMAGE_REGEX = /<image[^>]*>/i;
+
+let isValid = true;
```
paulbalaji, line 9: maybe this be an errorCount instead, that we increment

at the end we can check if > 0 and log number of errors found

scripts/optimize-svg.js
```diff
@@ -0,0 +1,112 @@
...
+        plugins: [
+          {
+            name: 'preset-default',
+            params: {
+              overrides: {
+                removeViewBox: false,
+              },
+            },
+          },
+          'removeScriptElement',
```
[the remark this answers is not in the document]
jmrossy: That's what this bit is referring to: https://www.notion.so/hyperlanexyz/Registry-review-process-and-issues-11a6d35200d6803499c4e0067ffbb45a?pvs=4#11b6d35200d680c9afe4c89d9757a8e7

scripts/optimize-svg.js
```diff
@@ -0,0 +1,112 @@
...
+        ],
+      });
+
+      if (result.error) {
+        console.error(`Error optimizing ${filePath}: ${result.error}`);
+        return; // Log the error and continue with the next file
+      }
```
jmrossy, line 93: Nit could you just `throw new Error(result.error`) here to avoid the duplicated log bit?

jmrossy (commented):
Mostly looks good! Visually the images look the same which is good. Can you confirm they're all smaller or roughly equal to what they were before?

scripts/optimize-svg.js
Xaroz: Done! Also changed with your previous comment of catching all the invalid paths and printing it at the end

.github/workflows/optimize-svg.yml
```diff
@@ -0,0 +1,57 @@
...
+          if [[ ! -z $CHANGES ]]; then
+            echo "Changes found: $CHANGES"
+            git diff
+            exit 1
+          fi
+
+      - name: setup-node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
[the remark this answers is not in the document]
Xaroz: done!

scripts/optimize-svg.js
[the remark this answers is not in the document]
Xaroz: Yeah, so `SVGO` by default in their `preset-default` configuration (https://svgo.dev/docs/preset-default/) already includes the plugin called: [removeUnknownsAndDefaults](https://svgo.dev/docs/plugins/removeUnknownsAndDefaults/), which removes any non-standard `SVG` tags (tested this with a few markups that are not `svgs`)

Xaroz:
> Mostly looks good! Visually the images look the same which is good. Can you confirm they're all smaller or roughly equal to what they were before?

I tested by using the same SVGs in the warp-ui template and they looked the same to me when comparing to live version (also manually checked a few of the SVGs)

xeno097 (approved):

jmrossy (approved):

jmrossy:
Jason and I discussed offline. He'll do a follow-up which separates the testing part and the optimization part so we can require tests to merge but allow failure of the file-writing parts, which will happen when a PR is from a 3rd party.

PR #454, opened 2024-12-23
feat: lower index chunks for some mostly non-EVM chains

tkporter:
### Description

- From https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4959#discussion_r1883786092


chains/sei/metadata.yaml
```diff
@@ -16,6 +16,8 @@ displayName: Sei
 domainId: 1329
 gasCurrencyCoinGeckoId: sei-network
 gnosisSafeTransactionServiceUrl: https://transaction.sei-safe.protofire.io
+index:
+  chunk: 1000
```
paulbalaji, line 20: 999 to err on the side of caution?

PR #455, opened 2024-12-26
chore: improve workflow

Xaroz:
### Description

Improve the workflow a bit with separating validation and optimization:
- Separate SVG validation and optimization
- Refactor files by making the access to get files path easier
- Add workflow for validation of files 

PD: still need to change optimization workflow to run on (probably) every commit to main

### Backward compatibility

Yes

### Testing

Manual


.changeset/fresh-turkeys-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Improve CI workflow by separating concerns
```
jmrossy, line 5: This is totally fine but FYI, changes that don't affect the published data don't need a changeset. Yes the changeset CI will fail but that's expected.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,46 @@
+name: validate-files
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
jmrossy, line 7: Why narrow this job down to these types?

scripts/common.js
jmrossy, line 1: Rename to utils, that's the convention we typically use

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,46 @@
...
+          if [[ ! -z $CHANGES ]]; then
+            echo "Changes found: $CHANGES"
+            git diff
+            exit 1
+          fi
+
+      - name: setup-node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20
```
jmrossy, line 38: Just curious, what happens if you remove this? I think the ubuntu image already has node. That's why the CI job doesn't include this step and why the `yarn install` step before it works.

.github/workflows/validate.yaml
Xaroz: The idea was to have the CI auto-run for every PR to the repo and this would prevent the CI to run on PRs that weren't ready. I know we discarded the idea of having it auto-run on every third-party PR, but I think this is still good, wdyt?

.github/workflows/validate.yaml
Xaroz: So doing a bit more digging, apparently the reason people still include this is that they don't want to rely on the default that comes installed with `ubuntu-latest` and have a bit more of a controlled environment. Other than that it also manages the Node.js installation in a way that is more optimized for Github Actions.

xeno097 (approved):

.github/workflows/validate.yaml
jmrossy: What's wrong with having it run on every PR?

.github/workflows/validate.yaml
jmrossy: I understand but we should be consistent. Either use it in all the workflows or none. And if you're going to add it, use `actions/setup-node@v4` which is the latest.

scripts/validate-file-path.js
```diff
@@ -0,0 +1,24 @@
+import { getFilePaths } from './utils.js';
```
jmrossy, line 1: FYI it doesn't have to be in this PR but this file would be a good place to solve this: https://github.com/hyperlane-xyz/hyperlane-registry/issues/209

.github/workflows/validate.yaml
Xaroz: The initial plan for this was that if we would change the permissions so that we wouldn't need to approve a workflow for it to run, so by doing these we would make sure only these types of PR would auto-run the workflow

.github/workflows/validate.yaml
Xaroz: Agreed here, I will remove this part for now and then look to add it back to every workflow after further discussions here

Xaroz: Agreed here, I will look into adding it to all the other workflows on another PR, we can discuss this further

scripts/validate-file-path.js
Xaroz: Sounds good, we can discuss this later, but I was maybe thinking having a different file for checks related to chain and checks related to deployments

.github/workflows/validate.yaml
jmrossy: We need to approve the workflows for security reasons, I don't think narrowing down the PR type will help. I recommend cutting this but it's up to you.

jmrossy (approved):

.github/workflows/validate.yaml
Xaroz: Removed PR types!

PR #456, opened 2024-12-31
feat: deploy sonicsvmtestnet, add solanatestnet addresses

tkporter:
### Description

- deploys `sonicsvmtestnet`
- adds addresses for that and `solanatestnet`


chains/sonicsvmtestnet/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+blocks:
+  confirmations: 1
+  estimateBlockTime: 0.4
+  reorgPeriod: 0
+chainId: 15153042
+deployer:
+  name: Abacus Works
+  url: https://www.hyperlane.xyz
+displayName: Sonic SVM Testnet
+displayNameShort: Sonic SVM Testnet
```
paulbalaji, line 16: don't need displayNameShort if it's just the same as displayName
```suggestion
```

paulbalaji (approved):

PR #457, opened 2025-01-02
chore: validate chains and warp routes, change jobs to run on main

### LINKED ISSUE #439: Improve registry review process with more checks
We’re getting a lot of PRs against the registry, both from 1st and 3rd parties. We need to distribute the review load and build tools + tests to automate some of the work. The current process + tooling creates more work for PR reviewers and submitters than is necessary. 

From: https://www.notion.so/hyperlanexyz/Registry-review-process-and-issues-11a6d35200d6803499c4e0067ffbb45a

Currently we don’t have coverage for these common mistakes which need to be caught manually

```[tasklist]
### Tasks
- [x] 1. Invalid or unsuitable chain logo file
- [x] 2. Incorrect file paths
- [x] 3. Invalid chain data
- [x] 4. Invalid Warp configs
- [ ] 5. Missing or invalid changeset level
- [x] 6. Run `combine` job and `optimize` job on every commit to main
```

Xaroz:
### Description

Fixes #439 and #209 

- Change `combine` and `optimize-svg` jobs to only run when commited to `main` 
- Add script to validate chains and warp routes
- Add `step` in `ci.ym`l test job to run these
- For chains it will check for missing `deployer` field and if its missing logo file
- For warp routes it will check for missing config file and if chain names are ordered correctly 
- Updated `chain` test to check if `isTestnet` is set properly
- Update `warp-routes` test to check for `logoURI` field

### Backward compatibility

Yes

### Testing

Manual


scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+function validateChains() {
+  if (!fs.existsSync(chainsDir)) {
+    console.error(`Directory does not exist: ${chainsDir}`);
+    return;
+  }
+
+  const entries = fs.readdirSync(chainsDir, { withFileTypes: true });
+
+  entries.forEach((entry) => {
+    if (entry.isDirectory()) {
```
jmrossy, line 27: Reduce nesting by inverting if statements like this.
`if (!entry.isDirectory()) return`

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+      const entryPath = path.join(chainsDir, entry.name);
+      const addressesPath = path.join(entryPath, 'addresses.yaml');
+      const metadataPath = path.join(entryPath, 'metadata.yaml');
+
+      if (fs.existsSync(metadataPath)) {
```
jmrossy, line 32: ditto

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+        // if addresses.yaml exists, check if deployer field is missing in metadata.yaml
+        const metadata = readYaml(metadataPath);
+        if (fs.existsSync(addressesPath)) {
+          if (!('deployer' in metadata)) {
```
jmrossy, line 36: It's nbd for a script like this but in general, avoid the `in` keyword due to complications with prototypal inheritance and use `Object.keys` (or `objKeys` from utils) instead.

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        // check if isTestNet is set properly for chains that could be testnets
+        // PD: this only works to check chains that have "test" on their name
+        if ('name' in metadata && metadata['name'].includes('test')) {
+          if (!('isTestnet' in metadata) || !metadata['isTestnet']) {
+            invalidTestnetChains.push(metadataPath);
+          }
+        }
+      }
```
jmrossy, line 48: IMO tests for file _contents_ should be with the existing unit tests (`test/unit/chains.test.ts`) that we we aren't splitting the logic into to places.

Tests for the file _structure_ (name, location, required neighboring files) can be here. 

WDYT?

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+
+// check if chain names in warp routes are ordered alphabetically
+function validateChainNameOrder(entryPath) {
+  const yamlFiles = fs.readdirSync(entryPath).filter((file) => file.includes('.yaml'));
+
+  yamlFiles.forEach((filename) => {
+    const match = filename.match(filenameRegex);
+    const filePath = path.join(entryPath, filename);
+
+    if (match) {
```
jmrossy, line 65: ditto nesting

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+  });
+}
+
+// This regex will make sure that we can split the filename when it contains the words
+// addresses or config that way we can grab all the chain names and exclude these words
+const filenameRegex = /^([\w-]+)-(addresses|config)\.yaml$/;
+
+// check if chain names in warp routes are ordered alphabetically
+function validateChainNameOrder(entryPath) {
```
jmrossy, line 58: This could maybe by simpler if you use `createWarpRouteConfigId` in `warp-utils` to generate the id and then compare it to the filename.

scripts/validate-file-data.js
```diff
@@ -0,0 +1,182 @@
...
+  const configFiles = fs.readdirSync(entryPath).filter((file) => file.includes('-config.yaml'));
+
+  if (configFiles.length === 0) {
+    noConfigFileError.push(entryPath);
+    return;
+  }
+
+  // Search and parse the config files in the current directory
+  // For each entry, find if field logoURI is missing
+  configFiles.forEach((configFile) => {
```
jmrossy, line 90: ditto, suggest moving to warp-routes.test.ts please

jmrossy (commented):
Very glad to have this new coverage but I think we should shuffle things around a bit (see comments)

scripts/validate-file-data.js
Xaroz: done

scripts/validate-file-data.js
Xaroz: We don't have the utils installed in the registry so I used Object.keys instead

scripts/validate-file-data.js
Xaroz: Moved test to chain.test.ts

scripts/validate-file-data.js
Xaroz: Same here!

scripts/validate-file-data.js
Xaroz: Our createWarpRouteConfigId is inside a ts script, so what I did is simply do:
 `const sortedChains = [...chains.split('-')].sort().join('-');`

inside the file

.github/workflows/validate.yml
```diff
@@ -2,7 +2,7 @@ name: validate-files
```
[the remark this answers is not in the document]
Xaroz: I moved the validates to inside the steps for the `test` job

scripts/validate-file-data.js
Xaroz: The initial reason I wanted to do this logic though it's that I wanted it to be `o(1)` best case, but I guess it doesn't really matter if this is more clean

scripts/validate-file-data.js
jmrossy: Feel free to add it as a dev dep if you want. Avoid using it as a runtime dep though.

jmrossy (approved):

scripts/validate-file-data.js
```diff
@@ -0,0 +1,131 @@
...
+    const logoFile = fs.readdirSync(entryPath).find((file) => file.includes('logo.svg'));
+    if (!logoFile) noLogoFileError.push(entryPath);
+
+    if (!fs.existsSync(metadataPath)) return;
+
+    // if addresses.yaml exists, check if deployer field is missing in metadata.yaml
+    const metadata = readYaml(metadataPath);
+    if (fs.existsSync(addressesPath)) {
+      if (!Object.keys(metadata).includes('deployer')) missingDeployerField.push(metadataPath);
+    }
```
xeno097, line 41: maybe in another pr instead of having to do this field manually checks you might want to create a zod schema for this so that we can do more checks also based on values of other fields. For example, if the deployer is abacus works then another field must be set too

xeno097 (approved):

scripts/validate-file-data.js
Xaroz: Makes sense, we can revisit this, I will add it to notion to keep a record of this

PR #34, opened 2024-05-17
Rename LocalRegistry to FileSystemRegistry

jmrossy:
### Description

Builds on https://github.com/hyperlane-xyz/hyperlane-registry/pull/33

Rename LocalRegistry to FileSystemRegistry to clarify its use case

### Backward compatibility

No

### Testing

Unit tests update

PR #464, opened 2025-01-07
Implement http registry

yorhodes:
### Description

- Implements `HttpServer` REST API wrapper around `IRegistry`
- Implements `HttpClientRegistry` which implements `IRegistry` for use in clients like the CLI or warp UI

### Next-steps

- Add `pino-http` for improve http logging, it has minimal dependencies, the largest being pino, which is already a dependency of the sdk so we should get dependency resolution if we ensure the versions are compatible, I've checked that the versions are compatible

### Backward compatibility

Yes

### Testing

Manual

## Summary by CodeRabbit

- **New Features**
  - Introduced support for HTTP-based registry services, allowing interaction with registries over HTTP.
  - Added an Express-powered HTTP server for registry access and management, including health check endpoints.
  - Provided new scripts for starting and developing the HTTP registry server.
  - Enabled remote registry interaction via a new HTTP client registry.
- **Bug Fixes**
  - Improved handling of registry URIs to distinguish between GitHub and generic HTTP registries.
- **Documentation**
  - Added internal documentation for new features and usage.
- **Chores**
  - Updated dependencies and configuration to support Express and TypeScript development workflows.


package.json
```diff
@@ -1,15 +1,17 @@
 {
   "name": "@hyperlane-xyz/registry",
   "description": "A collection of configs, artifacts, and schemas for Hyperlane",
-  "version": "6.11.0",
+  "version": "6.12.0-http.2",
   "dependencies": {
+    "express": "^4.21.2",
```
paulbalaji, line 6: wonder if it makes sense to spin this off to a side-package so it doesn't add to the normal registry size

yorhodes: alternatively I can do something similar to what we did with the `FileSystemRegistry` for tree shaking
https://github.com/hyperlane-xyz/hyperlane-registry/pull/34

src/registry/HttpClientRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  updateChain(update: UpdateChainParams): MaybePromise<void> {
+    return this.fetchJson<void>(`/chain/${update.chainName}/metadata`, {
+      method: 'POST',
+      body: JSON.stringify(update.metadata),
+    });
+  }
+
+  get type(): RegistryType {
```
jmrossy, line 37: Avoid `get` proxy methods. They have pitfalls. Use simple methods/members instead.

src/registry/HttpClientRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+    throw new Error('Method not implemented.');
```
jmrossy, line 38: Why not set a RegistryType?

src/registry/HttpClientRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  get uri(): string {
```
jmrossy, line 41: ditto

src/registry/HttpServer.ts
```diff
@@ -0,0 +1,81 @@
...
+  app: Express;
+
+  constructor(registry: IRegistry) {
+    this.app = express();
+    this.app.use(express.json());
+
+    this.app.get('/metadata', async (req, res) => {
+      try {
+        const metadata = await registry.getMetadata();
+        if (!metadata) {
```
jmrossy, line 15: Nbd but I don't think `getMetadata` can ever return a falsy value. Maybe you meant to check that `Object.keys().length > 0`?

src/registry/HttpServer.ts
```diff
@@ -0,0 +1,81 @@
...
+        return res.json(addresses);
+      } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : 'An unknown error occurred';
+        return res.status(500).json({ error: errorMessage });
+      }
+    });
+  }
+
+  start(port = process.env.PORT || 3000) {
+    return new Promise<void>((resolve, reject) => {
```
jmrossy, line 70: Does this method get `await`ed somewhere? Returning a process for a long-lived job like a server seems odd to me

jmrossy (commented):
This is cool!

src/registry/HttpServer.ts
yorhodes: this was AI generated lol
will clean this up before setting ready for review

src/registry/HttpServer.ts
yorhodes: also AI generated 😅

src/registry/HttpClientRegistry.ts
```diff
@@ -0,0 +1,124 @@
...
+    throw new Error('Method not implemented.');
+  }
+
+  getWarpRoute(routeId: string): MaybePromise<WarpCoreConfig | null> {
+    return this.fetchJson<WarpCoreConfig | null>(`/warp-route/${encodeURIComponent(routeId)}`);
+  }
+
+  getWarpRoutes(_filter?: WarpRouteFilterParams): MaybePromise<WarpRouteConfigMap> {
+    throw new Error('Method not implemented.');
+  }
```
Xaroz, line 84: is this not gonna be implemented?

.changeset/pre.json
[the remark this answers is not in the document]
Mo-Hussain: Whoops forgot to push a commit to clean this up, intention is for their to be one changeset

src/registry/HttpClientRegistry.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+  }
+
+  getChainAddresses(chainName: ChainName): MaybePromise<ChainAddresses | null> {
+    return this.fetchJson<ChainAddresses>(`/chain/${chainName}/addresses`);
+  }
+
+  getChainLogoUri(_chainName: ChainName): Promise<string | null> {
+    throw new Error('Method not implemented.');
+  }
```
xeno097, line 68: For the methods that won't be implemented, wouldn't it be better to log a warning instead of throwing so that the caller does not need to handle the exception or be aware of the type of registry used to avoid calling a method?

src/express/HttpServer.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.logger.info(`Server running on port ${port}`),
+      );
+
+      server.on('request', (req, _res) => this.logger.info('Request:', req.url));
+
+      server.on('error', (error) => this.logger.error('Server error:', error));
+    } catch (error) {
+      this.logger.error('Error starting server:', error);
+      process.exit(1);
+    }
```
xeno097, line 135: Can we have the endpoint methods implemented somewhere else as a function and register the endpoints and handlers using a method called `registerEndpoint` or something similar? It would make checking the endpoint implementations a bit easier and will make the `start` method smaller

src/express/HttpServer.ts
```diff
@@ -0,0 +1,137 @@
...
+const DEFAULT_PORT = 3000;
+const DEFAULT_REFRESH_INTERVAL = 1000 * 60 * 5; // 5 minutes
+
+export class HttpServer {
+  app: Express;
+  protected logger: Logger | Console;
+
+  constructor(protected getRegistry: () => Promise<IRegistry>, logger?: Logger) {
+    this.app = express();
+    this.app.use(express.json());
```
xeno097, line 17: We are not going to have any kind of auth on this server?

src/express/HttpServer.ts
```diff
@@ -0,0 +1,137 @@
...
+          const metadata = ChainMetadataSchema.parse(req.body);
+          await registry.updateChain({
+            chainName,
+            metadata,
+          });
+          return res.status(200).json({ message: 'Chain metadata updated successfully' });
+        } catch (error) {
+          const errorMessage = error instanceof Error ? error.message : 'An unknown error occurred';
+          return res.status(500).json({ error: errorMessage });
+        }
```
xeno097, line 94: You could implement an express middleware to catch the errors that happen in handlers so you wouldn't need to try catch each method because there would be a global errro handler.
ref: https://expressjs.com/en/guide/error-handling.html#:~:text=Writing%20error%20handlers

src/registry/HttpClientRegistry.ts
Mo-Hussain: We do follow this pattern in other `IRegistry` implementations https://github.com/hyperlane-xyz/hyperlane-registry/blob/2758d3adc22c3994a57758f955f0ee870477a229/src/registry/GithubRegistry.ts#L228

src/registry/HttpClientRegistry.ts
Mo-Hussain: I'll implement

src/express/HttpServer.ts
Mo-Hussain: The intetion is for this server to be run locally and as an internal service in our cluster so not needed

src/express/HttpServer.ts
Mo-Hussain: Added

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,30 @@
...
+
+interface ChainRequestParams {
+  chain: ChainName;
+}
+
+export function createChainRouter(chainService: ChainService) {
+  const router = Router();
+
+  router.get('/:chain/metadata', async (req: Request<ChainRequestParams>, res: Response) => {
+    const metadata = await chainService.getChainMetadata(req.params.chain);
```
xeno097, line 13: can check that the chain param is defined before sending it to the service?

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  router.post('/:chain/metadata', async (req: Request<ChainRequestParams>, res: Response) => {
+    const chain = ZChainName.parse(req.params.chain);
+    const _metadata = ChainMetadataSchema.parse(req.body);
+    const metadata = await chainService.setChainMetadata(chain, _metadata);
+    res.json(metadata);
+  });
+
+  router.get('/:chain/addresses', async (req: Request<ChainRequestParams>, res: Response) => {
+    const addresses = await chainService.getChainAddresses(req.params.chain);
```
xeno097, line 25: ditto

src/express/src/routes/warp.ts
```diff
@@ -0,0 +1,24 @@
...
+interface WarpRouteResponse {
+  data: WarpCoreConfig;
+}
+
+export function createWarpRouter(warpService: WarpService) {
+  const router = Router();
+
+  // get warp route
+  router.get('/:id', async (req: Request<WarpRouteParams>, res: Response<WarpRouteResponse>) => {
+    const warpRoute = await warpService.getWarpRoute(req.params.id);
```
xeno097, line 19: Can we check that the id is a string/defined?

src/registry/HttpClientRegistry.ts
xeno097: mmm should we keep doing that? i honestly would prefer a warning being logged than having to debug why something is not working only to discover that the underlying registry is throwing becase a method is throwing unexpectedly

src/express/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+    return next(err);
+  }
+
+  const apiError = err instanceof ApiError ? err : new ApiError('Internal Server Error', 500);
+  const status = apiError.status;
+  const body: ErrorResponse['error'] = {
+    message: apiError.message,
+  };
+
+  res.status(status).json(body);
```
xeno097, line 22: Shouldn't we log the error just in case something unexpected was thrown and we didn't wrap it in an `ApiError` ?

src/express/src/services/registryService.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.registry = await this.getRegistry();
+  }
+
+  async getCurrentRegistry(): Promise<IRegistry> {
+    const now = Date.now();
+    if (now - this.lastRefresh > this.refreshInterval) {
+      this.logger.info('Refreshing registry cache...');
+      this.registry = await this.getRegistry();
+      this.lastRefresh = now;
+    }
```
xeno097, line 25: Maybe it would be a good idea to add an or statement here in case someone forgets to call initialize? `(now - this.lastRefresh > this.refreshInterval) || !this.registry`

yarn.lock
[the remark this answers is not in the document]
Mo-Hussain: Also not clear to me why we have such a big diff, was an existing change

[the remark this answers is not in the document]
Mo-Hussain: @yorhodes - can you remember why?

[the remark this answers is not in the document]
yorhodes: adding express?

src/express/src/middleware/errorHandler.ts
yorhodes: agree we should log it on the server though the client will always get it in the body here iiuc

src/registry/HttpClientRegistry.ts
yorhodes: I think its pretty standard practice to throw behind a consistent interface with an unimplemented message
I do think we should be specific about which class does not have this behavior

src/express/src/routes/chain.ts
yorhodes: we should probably use a zod middleware or something for schema validation

src/fs/registry-utils.ts
```diff
@@ -71,14 +72,16 @@ export function getRegistry({
...
+      if (isHttpsUrl(uri) && uri.includes('github')) {
         return new GithubRegistry({
           uri,
           branch,
           logger: childLogger,
           proxyUrl: enableProxy ? PROXY_DEPLOYED_URL : undefined,
           authToken,
         });
+      } else if (uri.includes('http')) {
+        return new HttpClientRegistry(uri);
```
yorhodes, line 84: we should be careful to ensure that this usage does not imply the express dependency when tree-shaking

yarn.lock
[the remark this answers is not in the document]
Mo-Hussain: Cleaned this up

src/express/src/routes/chain.ts
Mo-Hussain: Example of the middleware in action 

```
curl -s -X POST "http://localhost:3000/chain/ethereum/metadata" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "ethereum",
    "chainId": 1,
    "domainId": 1,
    "protocol": "ethereum",
    "rpcUrls": [{"http": "https://eth-mainnet.g.alchemy.com/v2/your-api-key"}],
    "blockExplorers": [{"name": "Etherscan", "url": "https://etherscan.io"}],
    "blocks": {
      "confirmations": 1,
      "reorgPeriod": 64,
      "estimateBlockTime": 12
    }
  }' | jq '.'
{
  "message": "Validation error: blockExplorers.0.apiUrl: Required"
}
```

src/registry/HttpClientRegistry.ts
Mo-Hussain: I'll stick to throwing for now

src/express/src/middleware/errorHandler.ts
Mo-Hussain: Added server side logging

src/fs/registry-utils.ts
Mo-Hussain: `HttpClientRegistry` should be browser safe and `registry-utils.ts` is isolated properly in the fs module

xeno097 (approved):
Left just one last comment but approved on my side

package.json
```diff
@@ -12,6 +13,7 @@
     "@faker-js/faker": "^9.6.0",
     "@hyperlane-xyz/sdk": "12.3.0",
     "@types/chai-as-promised": "^8",
+    "@types/express": "^4",
```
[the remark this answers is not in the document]
Mo-Hussain: Good callout

antigremlin (commented):
Why are we putting the HTTP service into the same NPM Package as the client registry code? What are the pros and cons?

If this is the approach we choose, we need a few more bits:

1. Add a way to run the service locally, even if only via yarn.
2. Extend the restricted import rules in eslint to prevent accidental imports of service-related code.

package.json
```diff
@@ -38,6 +40,7 @@
     ".": "./dist/index.js",
     "./chains/schema.json": "./dist/chains/schema.json",
     "./chains/*": "./dist/chains/*",
+    "./express": "./dist/express/index.js",
```
antigremlin, line 43: What's the purpose of publishing the service code?

src/express/HttpServer.ts
```diff
@@ -0,0 +1,66 @@
...
+  async start(
+    port = parseInt(process.env.PORT || DEFAULT_PORT.toString()),
+    refreshInterval = parseInt(process.env.REFRESH_INTERVAL || DEFAULT_REFRESH_INTERVAL.toString()),
+  ) {
+    try {
+      const registryService = new RegistryService(this.getRegistry, refreshInterval, this.logger);
+      await registryService.initialize();
+
+      // add health check routes
+      this.app.use('/health', (_req: Request, res: Response) => void res.sendStatus(200));
```
antigremlin, line 35: Is this going to run in Kubernetes? If yes, I would seriously consider extracting the service into its own repo (or moving into the monorepo if this doesn't produce a circular dependency).

src/express/src/constants/ServerConstants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_PORT = 3000;
```
antigremlin, line 1: There seems to be no benefit in placing two constants into a separate file while they are only used once. The code would be easier to navigate if the constants were declared in the file where they are used.

By the way, do we want these values to be configurable?

src/express/index.ts
```diff
@@ -0,0 +1,2 @@
+// Exports for utilities that require express and are not suitable for browser use
+export { HttpServer } from './HttpServer.js';
```
antigremlin, line 2: Where is the code that actually _runs_ the HTTP service?

antigremlin (commented):
Here are a few questions from my side. I'll try to go over the rest of the code too.

For lack of a better place, it's rather questionable to have a directory structure like `src/express/src/stuff`.

package.json
Mo-Hussain: We are intending to run this server locally and as an internal service in our cluster, this is an example for how we would do this in the monorepo https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5120

src/express/index.ts
Mo-Hussain: I'll add this into the server to aid local development

src/express/src/constants/ServerConstants.ts
Mo-Hussain: I want to improve how we manage constants for things like the status codes by defining a class here. Both port and refresh interval can be passed in a arguments to the start method

src/express/HttpServer.ts
Mo-Hussain: I would not be opposed for the server to live in the monorepo. 
cc @yorhodes

src/express/src/middleware/validateRequest.ts
```diff
@@ -0,0 +1,35 @@
...
+import { z } from 'zod';
+import { ApiError } from '../errors/ApiError.js';
+
+type ValidationSchema = {
+  params?: z.ZodType;
+  query?: z.ZodType;
+  body?: z.ZodType;
+};
+
+export function validateRequest(schema: ValidationSchema) {
```
antigremlin, line 11: Validation might be more declarative and self-documenting if we rewrite this like so (I'm assuming the schemas are not async):

```typescript
export function validateRequestParam<T extends z.ZodTypeAny>(name: string, schema: T) {
  return (req: Request, res: Response, next: NextFunction) => {
    const parsed = schema.safeParse(req.params[name]);
    if (parsed.success) {
      req.params[name] = parsed.data;
      next();
    } else {
      res.status(404);
    }
  }
}

export function validateQueryParams<T extends z.ZodTypeAny>(schema: T) {
  return (req: Request, res: Response, next: NextFunction) => {
    const parsed = schema.safeParse(req.query);
    if (parsed.success) {
      req.query = parsed.data;
      next();
    } else {
      res.status(400).json(parsed.error.issues);
    }
  }
}

export function validateBody<T extends z.ZodTypeAny>(schema: T) {
  return (req: Request, res: Response, next: NextFunction) => {
    const parsed = schema.safeParse(req.body);
    if (parsed.success) {
      req.body = parsed.data;
      next();
    } else {
      res.status(400).json(parsed.error.issues);
    }
  }
}
```

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,45 @@
...
+const chainParamsSchema = z.object({
+  chain: ZChainName,
+});
+
+export function createChainRouter(chainService: ChainService) {
+  const router = Router();
+
+  router.get(
+    '/:chain/metadata',
+    validateRequest({ params: chainParamsSchema }),
```
antigremlin, line 16: And then

```suggestion
    validateRequestParam('chain', ZChainName),
```

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,45 @@
...
+      res.json(metadata);
+    },
+  );
+
+  router.post(
+    '/:chain/metadata',
+    validateRequest({
+      params: chainParamsSchema,
+      body: ChainMetadataSchema,
+    }),
```
antigremlin, line 28: And also

```suggestion
    validateRequestParam('chain', ZChainName),
    validateBody(ChainMetadataSchema),
```

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,45 @@
...
+    async (req: Request, res: Response) => {
+      const metadata = await chainService.setChainMetadata(req.params.chain, req.body);
+      res.json(metadata);
+    },
+  );
+
+  router.get(
+    '/:chain/addresses',
+    validateRequest({ params: chainParamsSchema }),
```
antigremlin, line 37:
```suggestion
    validateRequestParam('chain', ZChainName),
```

antigremlin (commented):
Here are some comments on validation

src/express/src/routes/root.ts
```diff
@@ -40,7 +37,8 @@ export function createRootRouter(rootService: RootService) {
   // get warp routes
   router.get(
     '/warp-routes',
-    validateRequest({ query: warpRoutesQuerySchema }),
+    validateQueryParam('symbol', ZSymbolOptional),
```
ltyu, line 40: since this is being passed into getWarpRoutes, shouldn't this match `WarpRouteFilterParams`? 

in otherwords, it doesnt quite match and can we make `WarpRouteFilterParams` a schema and just use that?

src/express/src/middleware/validateRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export function validateRequestParam<T extends z.ZodTypeAny>(name: string, schema: T) {
+  return (req: Request, _res: Response, next: NextFunction) => {
+    const parsed = schema.safeParse(req.params[name]);
+    if (parsed.success) {
+      req.params[name] = parsed.data;
+      next();
+    } else {
+      const errorMessage = parsed.error.errors
```
ltyu, line 33: should we use `parse()` and try-catch and ensure that this is a ZodError. 

otherwise some unknown, non-zod error may not have the exact parse.errors.erros structure.

src/express/src/middleware/validateRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+        new ApiError(
+          `Validation error in body: ${errorMessage}`,
+          AppConstants.HTTP_STATUS_BAD_REQUEST,
+        ),
+      );
+    }
+  };
+}
+
+export function validateQueryParam<T extends z.ZodTypeAny>(name: string, schema: T) {
```
ltyu, line 66: do we need all 3 of these functions? they seem pretty similar to each other

considering req.query, body, and params are all objects, we can let the schema validate and parse out what we need. could we have something like:
```
export function validateQueryParam<T extends z.ZodTypeAny>(obj: any, schema: T){
  schema.safeParse(obj)
  ...
  new ApiError(
}
```

You can then use 
```
const chainRequest = z.object({chain: string });
validateRequest(req.body, chainRequest);
validateRequest(req.params, chainRequest);
```

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,38 @@
...
+    validateRequestParam('chain', ZChainName),
+    async (req: Request, res: Response) => {
+      const metadata = await chainService.getChainMetadata(req.params.chain);
+      res.json(metadata);
+    },
+  );
+
+  router.post(
+    '/:chain/metadata',
+    validateRequestParam('chain', ZChainName),
```
ltyu, line 20: you can just include `:chain` within the body. i dont see any advantage of using both request params and body

src/express/src/routes/chain.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ChainMetadataSchema, ZChainName } from '@hyperlane-xyz/sdk';
+import { validateBody, validateRequestParam } from '../middleware/validateRequest.js';
+import { ChainService } from '../services/chainService.js';
+
+export function createChainRouter(chainService: ChainService) {
+  const router = Router();
+
+  router.get(
+    '/:chain/metadata',
+    validateRequestParam('chain', ZChainName),
```
ltyu, line 11: wouldn't it be easier to just use the queryParams?

src/express/src/routes/warp.ts
```diff
@@ -0,0 +1,24 @@
...
+const warpRouteParamsSchema = z.object({
+  id: z.string(),
+});
+
+export function createWarpRouter(warpService: WarpService) {
+  const router = Router();
+
+  // get warp route
+  router.get(
+    '/:id',
```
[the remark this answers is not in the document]
ltyu: would query params solve this issue?

ltyu (commented):
since we control the endpoint path, why do we need to request params? seems easier to just stick with query params for GET and body for POST

src/express/src/routes/chain.ts
xeno097: I'll play the devil's advocate but technically speaking, following the REST spec this is the correct way of defining  the endpoint

src/express/src/routes/warp.ts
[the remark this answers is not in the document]
Mo-Hussain: I think we will still need to do the encoding from the client

src/express/src/middleware/validateRequest.ts
Mo-Hussain: It's a mater of opinion but the declarative approach was suggested https://github.com/hyperlane-xyz/hyperlane-registry/pull/464#discussion_r2091760936

ltyu:
Closed in favor of the above merged PRs

PR #465, opened 2025-01-07
chore: update injective rpc URL

Xaroz:
### Description

Update Injective RPC URL

### Backward compatibility

Yes

### Testing

Manual


tkporter (approved):
ty

PR #472, opened 2025-01-08
fix: include PNDR logo URI

yorhodes:
### Description

Explicit logo path needed for warp UI?

### Backward compatibility

Yes

### Testing

Manual


Xaroz:
closing this as it is already added in #480

PR #499, opened 2025-01-17
fix: warp route artifact symbol

Xaroz:
### Description

This PR along with a PR to the monorepo aims to fix the artifacts for warp route deployments by taking into account the symbol field, mentioned in [this issue](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5164)

- Add `WarpRouteOptions` as optional parameter to `addWarpRoute`
- Change `getWarpRouteArtifactPaths` to receive a `symbol` as the folder path 
- Include unit test 
### Backward compatibility

Yes

### Testing

CLI


src/registry/IRegistry.ts
```diff
@@ -34,6 +34,10 @@ export enum RegistryType {
   Partial = 'partial',
 }
 
+export interface WarpRouteOptions {
```
jmrossy, line 37: This name feels to broad to me and could cause be confused with `WarpCoreConfig` and/or `WarpDeployConfig`. Let's call it some more specific to it's use like `AddWarpRouteParams` or `AddWarpRouteOptions`

jmrossy (commented):
LGTM just one comment

src/registry/IRegistry.ts
Xaroz: done!

jmrossy (approved):

.changeset/little-paws-begin.md
```diff
@@ -2,4 +2,4 @@
 '@hyperlane-xyz/registry': minor
 ---
 
-Add WarpRouteOptions to addWarpRoute function
+Include AddWarpRouteOptions to addWarpRoute function
```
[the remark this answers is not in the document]
Xaroz: > It's subtle but 'Include ... to' isn't quite right in English

😅 still more to learn

PR #513, opened 2025-01-22
refactor: consolidate warp addresses

mshojaei-txfusion:
### Description

Remove `-addresses.yaml` files for warp deployments since addresses are now maintained in config files. This change:
1. Removes warp-specific address YAML files
2. Removes warp address file generation functionality from registry
3. Updates registry tests to remove warp address file validation checks

This cleanup consolidates warp deployment addresses to a single location in the config files.

### Related issues
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5245

### Backward compatibility

Yes - This is a cleanup of redundant warp address files. All warp deployment addresses remain available in the config files.

### Testing

- Verified that all warp addresses previously in YAML files exist in corresponding config files
- Updated and ran registry tests to confirm functionality without warp address files
- Validated that CLI and other tooling continue to work with warp addresses from config files


nambrot (approved):
Noting that we'll need to update the CLI with this to avoid checking it in

test/unit/registry.test.ts
paulbalaji, line 1: could you pull latest main and remove any more addresses.yamls added in the interm and also add a test that checks that we _don't_ have an `addresses.yaml`?

mshojaei-txfusion: Sure, Done!

paulbalaji (approved):

PR #518, opened 2025-01-23
Add DeepBrainChain Mainnet

frank0528:
Description
Add DeepBrainChain Mainnet using the projects' official docs.

Backward compatibility
No

Testing
CLI


k4ctuss:
You need to deploy to deploy contracts on both chains with `hyperlane core deploy`, this will create you two files named addresses.yaml with the address of all contracts, see more [here](https://docs.hyperlane.xyz/docs/deploy-hyperlane#deploy-contracts)

PR #519, opened 2025-01-23
feat: jan 23 deploy batch

paulbalaji:
### Description

Deploy to abstract, glue, matchain, unitzero.

### Testing

manual


chains/abstract/metadata.yaml
```diff
@@ -0,0 +1,28 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://api.abscan.org/api
+    family: etherscan
+    name: Abstract Block Explorer
+    url: https://abscan.org
+blocks:
+  confirmations: 1
+  estimateBlockTime: 1
+  reorgPeriod: 0
```
tkporter, line 10: remind me where we landed w/ reorg periods for zksync chains?

tkporter (approved):

chains/abstract/metadata.yaml
paulbalaji: still 0 i believe, all the other zksync chains are still set to this too

PR #521, opened 2025-01-24
feat: warp speed warp configs reader

ltyu:

### DIFF
--- a/src/registry/FileSystemRegistry.ts
+++ b/src/registry/FileSystemRegistry.ts
@@ -5,7 +5,7 @@ import { parse as yamlParse } from 'yaml';
 
 import type { ChainMap, ChainMetadata, ChainName, WarpCoreConfig, WarpRouteDeployConfig } from '@hyperlane-xyz/sdk';
 
-import { CHAIN_FILE_REGEX, SCHEMA_REF, WARP_ROUTE_CONFIG_FILE_REGEX } from '../consts.js';
+import { CHAIN_FILE_REGEX, SCHEMA_REF, WARP_ROUTE_CONFIG_FILE_REGEX, WARP_ROUTE_DEPLOY_FILE_REGEX, } from '../consts.js';
 import { ChainAddresses, ChainAddressesSchema, WarpRouteId } from '../types.js';
 import { toYamlString } from '../utils.js';
 
@@ -18,7 +18,7 @@ import {
   type RegistryContent,
 } from './IRegistry.js';
 import { SynchronousRegistry } from './SynchronousRegistry.js';
-import { warpConfigToWarpAddresses, warpRouteConfigPathToId } from './warp-utils.js';
+import { warpConfigToWarpAddresses, warpRouteConfigPathToId, warpRouteDeployConfigPathToId } from './warp-utils.js';
 
 export interface FileSystemRegistryOptions {
   uri: string;
@@ -41,6 +41,9 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
     return path.join(this.uri, itemPath);
   }
 
+  /**
+   * Retrieves filepaths for chains, warp core, and warp deploy configs 
+   */
   listRegistryContent(): RegistryContent {
     if (this.listContentCache) return this.listContentCache;
 
@@ -63,7 +66,15 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
       warpRoutes[routeId] = filePath;
     }
 
-    return (this.listContentCache = { chains, deployments: { warpRoutes } });
+    const warpDeploys: RegistryContent['deployments']['warpDeploys'] = {};
+    const warpDeployFiles = this.listFiles(path.join(this.uri, this.getWarpRoutesPath()));
+    for (const filePath of warpDeployFiles) {
+      if (!WARP_ROUTE_DEPLOY_FILE_REGEX.test(filePath)) continue;
+      const routeId = warpRouteDeployConfigPathToId(filePath);
+      warpDeploys[routeId] = filePath;
+    }
+    
+    return (this.listContentCache = { chains, deployments: { warpRoutes, warpDeploys } });
   }
 
   getMetadata(): ChainMap<ChainMetadata> {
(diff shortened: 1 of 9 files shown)


src/registry/SynchronousRegistry.ts
```diff
@@ -72,9 +75,21 @@ export abstract class SynchronousRegistry extends BaseRegistry implements IRegis
     return Object.fromEntries(idsWithConfigs);
   }
 
+  /**
+   * Retrieves a map of all the warp routes deployment configs
+   */
+  getWarpDeploys() {
```
ltyu, line 81: Not sure if a filter parameter is needed yet

src/registry/SynchronousRegistry.ts
```diff
@@ -72,9 +75,21 @@ export abstract class SynchronousRegistry extends BaseRegistry implements IRegis
     return Object.fromEntries(idsWithConfigs);
   }
 
+  /**
+   * Retrieves a map of all the warp routes deployment configs
```
ltyu, line 79: Not sure if a filter parameter is needed yet

src/registry/FileSystemRegistry.ts
```diff
@@ -63,7 +66,15 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
...
-    return (this.listContentCache = { chains, deployments: { warpRoutes } });
+    const warpDeploys: RegistryContent['deployments']['warpDeploys'] = {};
+    const warpDeployFiles = this.listFiles(path.join(this.uri, this.getWarpRoutesPath()));
+    for (const filePath of warpDeployFiles) {
+      if (!WARP_ROUTE_DEPLOY_FILE_REGEX.test(filePath)) continue;
+      const routeId = warpRouteDeployConfigPathToId(filePath);
+      warpDeploys[routeId] = filePath;
+    }
+    
+    return (this.listContentCache = { chains, deployments: { warpRoutes, warpDeploys } });
```
jmrossy, line 77: Suggest renaming warpDeploys, it's not clear how it's different from wrapRoutes

src/registry/FileSystemRegistry.ts
```diff
@@ -182,9 +193,24 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
   }
 
   protected getWarpRoutesForIds(ids: WarpRouteId[]): WarpCoreConfig[] {
-    const configs: WarpCoreConfig[] = [];
     const warpRoutes = this.listRegistryContent().deployments.warpRoutes;
-    for (const [id, filePath] of Object.entries(warpRoutes)) {
+    return this.readConfigsForIds(ids, warpRoutes);
+  }
+
+  protected getWarpDeploysForIds(ids: WarpRouteId[]): WarpRouteDeployConfig[] {
```
jmrossy, line 200: ditto throughout

src/registry/IRegistry.ts
```diff
@@ -11,8 +11,11 @@ export interface RegistryContent {
   // Chain name to file type to file URI
   chains: ChainMap<ChainFiles>;
   deployments: {
-    // Warp route ID to config URI
+    // Warp route ID to artifact URI
     warpRoutes: Record<WarpRouteId, string>;
+
+    // Warp route ID to warp config URI
```
jmrossy, line 17:
```suggestion
    // Warp route ID to warp deploy config URI
```

src/registry/PartialRegistry.ts
```diff
@@ -58,6 +58,7 @@ export class PartialRegistry extends SynchronousRegistry implements IRegistry {
       chains,
       deployments: {
         warpRoutes,
+        warpDeploys: warpRoutes
```
jmrossy, line 61: intentional? Doesn't look right

src/registry/PartialRegistry.ts
```diff
@@ -87,6 +88,10 @@ export class PartialRegistry extends SynchronousRegistry implements IRegistry {
     }) as WarpCoreConfig[];
   }
 
+  protected getWarpDeploysForIds(_ids: WarpRouteId[]): WarpRouteDeployConfig[] {
+    throw new Error('Method not implemented.');
```
jmrossy, line 92: I'd expect this to filter and return some member var state like what `getWarpRoutesForIds` does

src/registry/SynchronousRegistry.ts
```diff
@@ -64,6 +64,9 @@ export abstract class SynchronousRegistry extends BaseRegistry implements IRegis
     return this.getWarpRoutesForIds([routeId])[0] || null;
   }
 
+  /**
+   * Retrieves a filtered map of the warp routes artifacts
```
jmrossy, line 68: Let's try to be consistent with our nomenclature. I don't think I call these artifacts anywhere in the registry code but I could be wrong. The problem with 'artifact' is that while it is the output of one tool, it's also the input to another.

src/registry/SynchronousRegistry.ts
jmrossy: Maybe not but I suggest including it for consistency. If you don't do it now, it'll probably be 6 months before this is touched again

test/unit/warp-routes.test.ts
```diff
@@ -19,15 +19,15 @@ describe('Warp Route Configs', () => {
   const routes = localRegistry.getWarpRoutes();
 
   for (const id of Object.keys(routes)) {
-    it(`Route ${id} is valid`, async () => {
+    it(`Core Config ${id} is valid`, async () => {
```
jmrossy, line 22: Replace Core with WarpCore. Core is too ambiguous with hyperlane core config IMO

test/unit/warp-routes.test.ts
```diff
@@ -120,4 +120,26 @@ describe('Warp Route Configs', () => {
...
+    'TIA/eclipsemainnet-stride', 
+    'TIA/mantapacific-neutron', 
+    'stTIA/eclipsemainnet-stride'
+  ];
+  const configs = Object.keys(warpDeploys).filter(id =>!excludeIds.includes(id));
+  for (const id of configs) {
+    it(`Deploy config ${id} is valid`, async () => {
+      WarpRouteDeployConfigSchema.parse(warpDeploys[id]);
+    });
+  }
```
jmrossy, line 144: Please put this section in a separate `describe` block

jmrossy (commented):
Seems like you're on the right track

src/registry/PartialRegistry.ts
ltyu: This is kind of hard to implement because the warpRouteId includes the symbol name (`USDC/base-ethereum`), which WarpDeployConfig does not have. We will need to derive from the `config.token`.

Is it worth implementing now as it seems like we may merge both configs soon

jmrossy: Ah, k nvm

jmrossy: Please leave a comment about this though

src/registry/SynchronousRegistry.ts
ltyu: Added

src/registry/FileSystemRegistry.ts
```diff
@@ -63,7 +66,15 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
...
-    return (this.listContentCache = { chains, deployments: { warpRoutes } });
+    const warpDeployConfigURIs: RegistryContent['deployments']['warpDeployConfigURIs'] = {};
+    const warpDeployFiles = this.listFiles(path.join(this.uri, this.getWarpRoutesPath()));
+    for (const filePath of warpDeployFiles) {
+      if (!WARP_ROUTE_DEPLOY_FILE_REGEX.test(filePath)) continue;
+      const routeId = warpRouteDeployConfigPathToId(filePath);
+      warpDeployConfigURIs[routeId] = filePath;
+    }
+    
+    return (this.listContentCache = { chains, deployments: { warpRoutes, warpDeployConfigURIs } });
```
jmrossy, line 77: Nit: wondering why the `warpDeployConfigURIs` name feels inconsistent with the other exports here. Would `warpDeployConfigs` be better?

jmrossy (approved):

src/registry/FileSystemRegistry.ts
ltyu: true. Updated

PR #569, opened 2025-02-08
Add bouncebitmainnet chain

LyDe17TC:
added bouncebitmainnet chain, 
the logo has been minimized with the recommended application and the metadata comes from the api


paulbalaji:
done as part of https://github.com/hyperlane-xyz/hyperlane-registry/pull/596

LyDe17TC:
ok good job

PR #572, opened 2025-02-09
chore: fix incorrect variable check in routing logic

0xminds:
### Description

I noticed that the condition was checking the `merkle_tree_hook` variable, but it should have been checking `fallback_routing_hook`.

I've updated the code to reflect the correct variable.


0xminds:
[Mo-Hussain](https://github.com/Mo-Hussain), please check it bro!

PR #586, opened 2025-02-13
chore: add drpc endpoint to default ethereum RPC set

paulbalaji:
### Description

chore: add drpc endpoint to default ethereum RPC set


ltyu (approved):
just update wss

PR #610, opened 2025-02-20
feat: authenticated GitHub registry

mshojaei-txfusion:
### Description

Add authentication support for GitHub registry access. This enhancement allows the registry to make authenticated requests to GitHub's API using a personal access token, which helps avoid rate limiting issues and enables access to private repositories. The implementation includes:

- Added `authToken` option to GithubRegistry configuration
- Updated API request headers to include authentication when token is provided
- Added test environment configuration
- Enhanced test coverage for authenticated scenarios
- Added proper error handling for API rate limits

### Related Issues
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5501

### Backward compatibility

Yes. This change is fully backward compatible as:
1. The `authToken` parameter is optional
2. The existing functionality remains unchanged when no authentication is provided
3. All existing methods maintain their current signatures and behavior

### Testing

- Added new unit tests to verify authenticated API calls
- Tested rate limit handling with and without authentication
- Verified existing registry functionality remains intact
- All tests passing in the test environment with both authenticated and unauthenticated scenarios


src/registry/GithubRegistry.ts
```diff
@@ -249,7 +252,12 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
 
   protected async fetch(url: string): Promise<Response> {
     this.logger.debug(`Fetching from github: ${url}`);
-    const response = await fetch(url);
+    const useToken = !(this.proxyUrl && url.startsWith(this.proxyUrl)) && !!this.authToken;
```
nambrot, line 255: Optional Nit: This is a quite hard to reason about condition, consider renaming some of the subconditions, something like `!proxied & hasToken`

src/registry/GithubRegistry.ts
```diff
@@ -249,7 +252,12 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
...
+    const response = await fetch(url, {
```
nambrot, line 256: I just note this now and maybe I'm too boomer, but do we need a polyfill for fetch for usage on the frontend?

nambrot (approved):
Seems simple enough, but can we make sure that when we release this to test this manually well? @ltyu @cmcewen

src/registry/GithubRegistry.ts
cmcewen: https://caniuse.com/fetch

cmcewen:
@nambrot could you be more specific about what you want for testing?

nambrot:
I just want to make sure that when we merge this into `main`, that we test that the normal usage (i.e. without tokens or private registries) continues to work. I believe it does, but I'm not confident that our automated tests would cover this rn

paulbalaji:
changes lgtm but wondering why there's 1 test that says "pending"? https://github.com/hyperlane-xyz/hyperlane-registry/actions/runs/13817132074/job/38653269597?pr=610#step:4:5494
image

paulbalaji (approved):

mshojaei-txfusion:
> changes lgtm but wondering why there's 1 test that says "pending"? https://github.com/hyperlane-xyz/hyperlane-registry/actions/runs/13817132074/job/38653269597?pr=610#step:4:5494 image
> 
> I see the same locally with `yarn test:unit` image

This for a test case that requires github toke to pass, otherwise will skip it:

image

PR #630, opened 2025-02-28
feat: Add custom ESLint plugin for YAML sorting

mshojaei-txfusion:
### Description

Added a custom ESLint plugin that enforces consistent sorting in YAML files. The plugin specifically targets arrays in chain and deployment configuration files, sorting:
- `tokens` arrays by `chainName`
- `tokens[].connections` arrays by `token`

This improves readability, reduces merge conflicts, and makes changes to these arrays more predictable and consistent.

Also added a convenience npm script `sort-yaml` to easily format all YAML files.

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5547

### Backward compatibility

Yes. This change only affects linting rules and doesn't modify any runtime behavior.

### Testing

The plugin has been tested against existing YAML files in the `chains` and `deployments` directories. The sorting logic properly preserves all data while reorganizing it consistently.


chains/alephzeroevmmainnet/metadata.yaml
```diff
@@ -24,6 +24,6 @@ nativeToken:
   symbol: AZERO
 protocol: ethereum
 rpcUrls:
-  - http: https://rpc.alephzero.raas.gelato.cloud
   - http: https://alephzero.drpc.org
+  - http: https://rpc.alephzero.raas.gelato.cloud
```
nambrot, line 28: We should exclude RPC urls from sorting (because there might be a reason to sort these according to reliability

mshojaei-txfusion: Good catch, I was not aware of it.
fixed also did same for block explorers

mshojaei-txfusion:
what worries me is this PR removes yaml files comments.
@ltyu can you please take a look at PR first then let me know if it worth it working on this comments thing or not

ltyu:
> what worries me is this PR removes yaml files comments. @ltyu can you please take a look at PR first then let me know if it worth it working on this comments thing or not

Ah i actually didnt submit my review and was wondering why they were removed

we'd prefer to keep the comments, especially the top `yaml-language-server...` because it's used for schema validation iirc

package.json
```diff
@@ -70,7 +70,8 @@
     "test:unit": "yarn build && mocha --config .mocharc.json './test/unit/*.test.ts' --exit",
     "test:health": "yarn build && mocha --config .mocharc.json './test/health/*.test.ts' --exit --parallel",
     "version:prepare": "yarn changeset version && yarn install --no-immutable",
-    "version:check": "yarn changeset status"
+    "version:check": "yarn changeset status",
+    "sort-yaml": "eslint --fix chains/**/*.yaml deployments/**/*.yaml"
```
ltyu, line 74: i believe this only lints when calling `yarn sort-yaml`, right?

ltyu:
> what worries me is this PR removes yaml files comments. @ltyu can you please take a look at PR first then let me know if it worth it working on this comments thing or not

any updates on how to keep the comments?

package.json
ltyu: would love some way to run this automatically, or enforce it. Perhaps using an Action similar to what we do in monorepo https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/.github/workflows/test.yml#L80-L87

mshojaei-txfusion:
> > what worries me is this PR removes yaml files comments. @ltyu can you please take a look at PR first then let me know if it worth it working on this comments thing or not
> 
> any updates on how to keep the comments?

Yes working on it and almost finished, but as a plugin in monorepo utils, so we can use it here

mshojaei-txfusion:
> > what worries me is this PR removes yaml files comments. @ltyu can you please take a look at PR first then let me know if it worth it working on this comments thing or not
> 
> any updates on how to keep the comments?

Please check this out:
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5909
finally I I managed to keep comments manually

mshojaei-txfusion:
As the Eslint plugin moved into utils package, closed this and opened: https://github.com/hyperlane-xyz/hyperlane-registry/pull/887

PR #636, opened 2025-03-04
feat: file path validation

mshojaei-txfusion:
### Description

Add file path validation for FileSystemRegistry. This implementation validates paths provided to the registry by checking for invalid characters, ensuring URL-like paths use the `file:// protocol`, and detecting null bytes that could be used in path traversal attacks. The validation throws a clear error message when invalid paths are detected, preventing potential runtime errors or security issues.

### Backward compatibility

Yes - This change is backward compatible as it only adds validation to reject invalid paths that would likely cause errors during operation anyway. Valid paths will continue to work as before.

### Testing

The validation function has been manually tested with various path formats to ensure proper rejection of invalid paths and acceptance of valid ones. The implementation works correctly with existing registry functionality.


src/registry/registry-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const isValidFilePath = (path: string): boolean => {
+  try {
+    // Check for invalid characters in path
+    // These are generally invalid in most file systems
+    const invalidChars = /[<>:"|?*\x00-\x1F]/g;
+    if (invalidChars.test(path)) return false;
+
+    // For paths with protocol, validate they're file:// protocol
+    if (path.includes('://')) {
```
nambrot, line 30: If we are going to support `file://` strings, then we need to actually extract the path for file system registry, no?

mshojaei-txfusion: Right, missed it and did not know file registry just handles file system file paths, fixed.

mshojaei-txfusion:
> lgtm, can you just make sure we don't break the dependencies again by using "wrong" dependencies?

Tested on `hyperlane-warp-ui-template` with `yarn link` build is fine with this branch.

PR #649, opened 2025-03-06
feat: march 5 testnet deploys

paulbalaji:
### Description

- feat: deploy to carrchaintestnet, somniatestnet, infinityvmmonza
- adds metadata for megaethtestnet, rometestnet

### Testing

manual


chains/infinityvmmonza/metadata.yaml
```diff
@@ -0,0 +1,23 @@
+# yaml-language-server: $schema=../schema.json
```
Mo-Hussain, line 1: Assuming it is ok not have block explorer defined?

Mo-Hussain (approved):

chains/infinityvmmonza/metadata.yaml
paulbalaji: yup, believe there's precedence for this with other testnets e.g. eclipse testnet

PR #656, opened 2025-03-07
feat: Add parseGitHubPath() to parse URI in GithubRegistry

ltyu:
### Description
Update GithubRegistry to be able to parse branch from URI by using a helper function. The function uses regex instead of split/positional indices because we may have branches that look like: `ltyu/REZ/base-ethereum`

### Drive-by
- Remove `isCanonicalRepoUrl()`
- Some lint updates.

### Backward compatibility
Yes - moved the existing parsing logic to a helper function that uses regex instead of positional arrays.

### Testing
Unit test


test/unit/registry.test.ts
```diff
@@ -331,12 +332,12 @@ describe('Registry Utils', () => {
...
       {
         name: 'mixed registry types with proxy settings',
-        uris: [DEFAULT_GITHUB_REGISTRY, localPath, 'https://github.com/test'],
+        uris: [DEFAULT_GITHUB_REGISTRY, localPath, 'https://github.com/user/test'],
         useProxy: true,
         expectedRegistries: [
           { type: GithubRegistry, uri: DEFAULT_GITHUB_REGISTRY, proxyUrl: PROXY_DEPLOYED_URL },
           { type: FileSystemRegistry, uri: localPath },
-          { type: GithubRegistry, uri: 'https://github.com/test' },
+          { type: GithubRegistry, uri: 'https://github.com/user/test' },
```
nambrot, line 340: Can you add a test case here that includes the branch in the URL?

src/registry/registry-utils.ts
```diff
@@ -61,7 +57,7 @@ export function getRegistry(
           uri,
           branch,
           logger: childLogger,
-          proxyUrl: enableProxy && isCanonicalRepoUrl(uri) ? PROXY_DEPLOYED_URL : undefined,
+          proxyUrl: enableProxy ? PROXY_DEPLOYED_URL : undefined,
```
ltyu, line 60: The intention of isCanonicalRepoUrl() was for extra protection by preventing callers that enable proxy but use a non-hyperlane registry url. This extra "protection" is unnecessary.

src/registry/GithubRegistry.ts
```diff
@@ -67,11 +77,12 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
...
-    const pathSegments = this.url.pathname.split('/');
-    if (pathSegments.length < 2) throw new Error('Invalid github url');
-    this.repoOwner = pathSegments.at(-2)!;
-    this.repoName = pathSegments.at(-1)!;
+
+    const { repoOwner, repoName, repoBranch } = parseGitHubPath(this.uri);
+
+    this.repoOwner = repoOwner;
+    this.repoName = repoName;
+    this.branch = options.branch ?? repoBranch ?? 'main';
```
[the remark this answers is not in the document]
ltyu: Yup added

nambrot (approved):

src/registry/registry-utils.ts
nambrot: Does the proxy not work for arbitrary repos? If so why not?

ltyu: design decision to prevent others from using our API key to make arbitrary requests

PR #664, opened 2025-03-11
add dravana network

pullstacker:
### Description

Add Dravana layer3 mainnet

### Backward compatibility

Yes

### Testing

Yes we have tested by using Hyperlane CLI


JaunePomme:
Overall looks good to me .. run yarn prettier and lint just in case, may add your name in a subfield "deployer" in your .yml, minimify with https://jakearchibald.github.io/svgomg/  if you haven't already
And then I think it's fine

PR #666, opened 2025-03-11
Feat/get warp deploy path

mshojaei-txfusion:
### Description

Add `getWarpDeployArtifactPaths` static function to BaseRegistry and improve warp route config ID generation with optional symbol parameter. This change enhances the flexibility of warp route configuration handling and provides better support for deploy artifact path generation.

Key changes:
- Add new static `getWarpDeployArtifactPaths` function to BaseRegistry
- Improve warp route config ID generation with optional symbol parameter
- Refactor warp route and deploy config artifact path generation for better maintainability
- Update build script to support the new warp route config ID generation

### Releated issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5244

Related to ENG-1014

### Backward compatibility

Yes

These changes are backward compatible as they:
1. Add new functionality without removing existing features
2. Maintain the same config file structure and naming conventions
3. Keep existing function signatures while adding optional parameters

### Testing

The changes have been tested through:
- Build script execution with existing warp configs


src/registry/BaseRegistry.ts
```diff
@@ -47,20 +54,24 @@ export abstract class BaseRegistry implements IRegistry {
...
+  }
+
+  static getWarpDeployArtifactPaths(
+    deployConfig: WarpDeployConfigMap,
+    symbol: string,
+    configBasePath: string = './configs',
+  ) {
+    const chains = Object.keys(deployConfig);
+    return {
+      configPath: `${configBasePath}/${createWarpRouteConfigId(symbol, chains)}-deploy-config.yaml`,
```
antigremlin, line 73: Should it be `-deploy.yaml` instead of `-deploy-config.yaml`?

antigremlin (commented):
This new function intends to mirror the existing `getWarpRoutesArtifactPaths` but it doesn't exactly do that. There should be good reasons for the differences like:

- having a static function vs. protected
- having `configBasePath` as an argument vs. calling `this.warpRoutesPath()`

It seems these differences stem from the fact that no code uses this function so far. The following steps may help improve the design:

1. Start calling `getWarpDeployArtifactPaths` from `addWarpRouteConfig` to mirror the way `getWarpRoutesArtifactPaths` is used. This probably breaks the `export-warp-configs.ts` script that uses the function. To avoid immediate breakage, we can introduce an overload (or another function with a similar name) so that we can change the code using this and remove the older variant later.
2. Introduce unit tests for `getWarpRoutesArtifactPaths` and `getWarpDeployArtifactPaths`.

src/registry/BaseRegistry.ts
mshojaei-txfusion: What you say make sense, but this was what @nambrot asked on isseu: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5244

any reason @nambrot ?

src/registry/BaseRegistry.ts
```diff
@@ -47,20 +54,24 @@ export abstract class BaseRegistry implements IRegistry {
...
+      )}-config.yaml`,
+    };
+  }
+
+  static getWarpDeployArtifactPaths(
+    deployConfig: WarpDeployConfigMap,
+    symbol: string,
+    configBasePath: string = './configs',
+  ) {
+    const chains = Object.keys(deployConfig);
```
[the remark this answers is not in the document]
mshojaei-txfusion: This is supposed to call for warp init output file path, so we need it to be static

mshojaei-txfusion:
> This new function intends to mirror the existing `getWarpRoutesArtifactPaths` but it doesn't exactly do that. There should be good reasons for the differences like:
> 
> * having a static function vs. protected
> * having `configBasePath` as an argument vs. calling `this.warpRoutesPath()`
> 
> It seems these differences stem from the fact that no code uses this function so far. The following steps may help improve the design:
> 
> 1. Start calling `getWarpDeployArtifactPaths` from `addWarpRouteConfig` to mirror the way `getWarpRoutesArtifactPaths` is used. This probably breaks the `export-warp-configs.ts` script that uses the function. To avoid immediate breakage, we can introduce an overload (or another function with a similar name) so that we can change the code using this and remove the older variant later.
> 2. Introduce unit tests for `getWarpRoutesArtifactPaths` and `getWarpDeployArtifactPaths`.

This would be used as a replacement for default output path on `createWarpRouteDeployConfig`.
overloading is good idea, will do it.

mshojaei-txfusion:
> This new function intends to mirror the existing `getWarpRoutesArtifactPaths` but it doesn't exactly do that. There should be good reasons for the differences like:
> 
> * having a static function vs. protected
> * having `configBasePath` as an argument vs. calling `this.warpRoutesPath()`
> 
> It seems these differences stem from the fact that no code uses this function so far. The following steps may help improve the design:
> 
> 1. Start calling `getWarpDeployArtifactPaths` from `addWarpRouteConfig` to mirror the way `getWarpRoutesArtifactPaths` is used. This probably breaks the `export-warp-configs.ts` script that uses the function. To avoid immediate breakage, we can introduce an overload (or another function with a similar name) so that we can change the code using this and remove the older variant later.
> 2. Introduce unit tests for `getWarpRoutesArtifactPaths` and `getWarpDeployArtifactPaths`.

After warp deployment we use `registry.addWarpRoute` to store it on registry, so we do not need access path outside of registry as we write there.

But this is not cleared to me yet, are we migrating to use `addWarpRouteConfig` on CLI?

@nambrot can you please clarify this about issue created, because if this is the case, what @antigremlin said totally makes sense.

.changeset/silent-items-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': minor
+---
+
+Add `getWarpDeployArtifactPaths` funciton to BaseRegistry
```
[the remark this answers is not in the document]
mshojaei-txfusion: resolved

src/registry/warp-utils.ts
```diff
@@ -64,9 +64,15 @@ function parseWarpRouteConfigPath(configRelativePath: string, regex: RegExp) {
...
  * symbols are not officially supported yet.
  */
-export function warpRouteConfigToId(config: WarpCoreConfig): WarpRouteId {
+export function warpRouteConfigToId(config: WarpCoreConfig, symbol?: string): WarpRouteId {
   if (!config?.tokens?.length) throw new Error('Cannot generate ID for empty warp config');
-  const tokenSymbol = config.tokens[0].symbol;
+  const symbols = new Set<string>(config.tokens.map((token) => token.symbol.toUpperCase()));
+  if (!symbol && symbols.size !== 1) {
+    throw new Error(
+      'Only one token symbol per warp config is supported for now. Consider passing a symbol as a parameter',
```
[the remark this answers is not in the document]
mshojaei-txfusion: Good idea, resolved.

src/registry/BaseRegistry.ts
paulbalaji: we already have `-config.yaml` so adding a `-deploy-config.yaml` might add confusion

mshojaei-txfusion: Yes make sense

antigremlin (approved):
Looks good with all the recent changes

src/registry/FileSystemRegistry.ts
```diff
@@ -119,13 +120,29 @@ export class FileSystemRegistry extends SynchronousRegistry implements IRegistry
...
-    this.createFile({ filePath: configPath, data: toYamlString(config, SCHEMA_REF) });
+    const configPath = this.getWarpRouteCoreConfigPath(config, options);
+    this.createFile({
+      filePath: path.join(this.uri, configPath),
+      data: toYamlString(config, SCHEMA_REF),
+    });
   }
+  //TODO: This string parameter overload is for backwards compatibility with the export-warp-configs.ts script.
+  //It should be removed when all consumers have been updated to use the options parameter.
+  //See: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/eb3054c59184573f67f79a801965c8e4cc2ed3ce/typescript/infra/scripts/warp-routes/export-warp-configs.ts#L35
```
ltyu, line 131: Can we just change export-warp-configs.ts instead of doing this?

ltyu: fine to do this in separate issue

ltyu (approved):

src/registry/BaseRegistry.ts
```diff
@@ -47,20 +55,24 @@ export abstract class BaseRegistry implements IRegistry {
...
-    const basePath = `${this.getWarpRoutesPath()}/${symbol}/${chains}`;
-    return { configPath: `${basePath}-config.yaml` };
+  protected getWarpRouteCoreConfigPath(config: WarpCoreConfig, options?: AddWarpRouteOptions) {
+    return `${this.getWarpRoutesPath()}/${warpRouteConfigToId(
+      config,
+      options?.symbol,
+    )}-config.yaml`;
+  }
+
+  protected getWarpRouteDeployConfigPath(
```
ltyu, line 65: can we get a unit test for these 2 fns?

src/registry/FileSystemRegistry.ts
mshojaei-txfusion: will  do it after we released and had access to this function there.

src/registry/BaseRegistry.ts
mshojaei-txfusion: Done, please check and merge.

PR #668, opened 2025-03-12
feat: add Turtle

christopherbrumm:
### Description

Add Turtle chain warp route from Ethereum (collateral) to Linea (synthetic).

### Backward compatibility

Yes

### Testing

- Warp checker
- Contracts are verified


deployments/warp_routes/TURTLE/logo.svg
paulbalaji, line 1: Seems quite large for an svg, are there any smaller sized pictures we can use? if you can find a png that's smaller then that is also acceptable for warp routes (it's only chain-logos where we specifically demand svgs)

ltyu (approved):

deployments/warp_routes/TURTLE/logo.svg
christopherbrumm: Updated ✅

paulbalaji (approved):

PR #688, opened 2025-03-19
chore: remove total supply from deployments config

Xaroz:
### Description

Remove totalSupply from token metadata in warp deployments. The field seems to not be needed in most cases and could cause misconfiguration issues, on the synthetic token contract `totalSupply` was used, which mints tokens to the message sender. 

Instead we now apply a `initialSupply` directly to the synthetic token without the need to use `totalSupply`

Related [PR](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5686) from the monorepo
### Backward compatibility

Yes

### DIFF
--- a/.changeset/slimy-moose-lick.md
+++ b/.changeset/slimy-moose-lick.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': minor
+---
+
+Remove totalSupply from warp route deployments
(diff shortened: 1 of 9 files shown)


paulbalaji:
what's the context for this change?

paulbalaji (approved):

paulbalaji:
can we also add a test to make sure there's no `totalSupply` in future configs?

ltyu (approved):

PR #698, opened 2025-03-25
fix: set forma IGP to zero address

paulbalaji:
### Description

fix: set forma IGP to zero address
- so scraper doesn't complain


tkporter (approved):

tkporter (changes requested):
actually should probably index the existing Forma IGP

tkporter (approved):
chatted offline, seems no igp is present

PR #728, opened 2025-04-02
feat: update kyvetestnet protocol type to `cosmosnative`

paulbalaji:
### Description

Update kyvetestnet protocol from cosmos to cosmosnative.
Drive-by yarn build.


chains/addresses.yaml
```diff
@@ -45,7 +45,7 @@ alephzeroevmmainnet:
   interchainAccountIsm: "0x4d264424905535E97396Db83bd553D0d73A4EF9d"
   interchainAccountRouter: "0x26A29486480BD74f9B830a9B8dB33cb43C40f496"
   interchainGasPaymaster: "0x9c2214467Daf9e2e1F45b36d08ce0b9C65BFeA88"
-  interchainSecurityModule: "0x75177B4869d1DD30A1717db94d999cBe5C228211"
+  interchainSecurityModule: "0xAF6B4222dB762a21E5DB86e5a096A9f61ff8536B"
```
tkporter, line 48: why all these changes?

tkporter (approved):

chains/kyvetestnet/metadata.yaml
```diff
@@ -33,7 +33,7 @@ nativeToken:
   denom: tkyve
   name: KYVE
   symbol: KYVE
```
tkporter, line 35: kinda wonder if this should've been named something to do with kaon, but nbd

chains/addresses.yaml
paulbalaji: we locked down some repo settings so the GH action that does these updates on main automatically is not fully working - the `yarn build` updates the concatenated addresses/metadata file and that's not run since merging the 2 recent ISM update PRs

chains/kyvetestnet/metadata.yaml
paulbalaji: i think the network is still called kyve kaon so idm keeping the nativetoken as kyve (which is what the explorer shows)

chains/addresses.yaml
paulbalaji: no longer necessary now that CI is fixed - removing from PR to cleanup diff

PR #729, opened 2025-04-02
chore(ci): split nitro check+update to a separate job

paulbalaji:
### Description

We're currently spending a lot of time laboriously checking if a chain is a nitro chain in main before doing the metadata/addresses update. This PR splits the check to a separate job so that the actual built-out metadata/addresses files can be updated quicker.

### Backward compatibility

yes

### Testing

ci

### DIFF
--- a/.github/workflows/update-nitro-chains.yml
+++ b/.github/workflows/update-nitro-chains.yml
@@ -1,4 +1,4 @@
-name: combine
+name: update-nitro-chains
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


Xaroz (approved):
approved, but how often does the update-nitro-chain happens? There could be a possibility of optimize svg, combine and update-nitro-chains happening thus triggering three different actions

paulbalaji:
> approved, but how often does the `update-nitro-chain` happens? There could be a possibility of `optimize-svg`, `combine` and `update-nitro-chains` happening thus triggering three different actions, not sure if we want this

very rarely - it's typically caught at review time. atm we already have scenarios where the optimise+combine jobs update main one after the other e.g. 
image

PR #748, opened 2025-04-10
chore(ci): parallelise ci

paulbalaji:
### Description

chore(ci): parallelise ci
- we don't really need to wait for each step to complete before kicking off the next one, so let's just do it all in parallel

### Testing

[Before](https://github.com/hyperlane-xyz/hyperlane-registry/actions/runs/14380392593?pr=748)
Before

[After](https://github.com/hyperlane-xyz/hyperlane-registry/actions/runs/14380294352?pr=746)
After

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,6 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    needs: [install]
     steps:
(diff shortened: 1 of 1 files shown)


Xaroz (approved):
LGTM, but your Before links to the after CI job and the after links to the previous

PR #758, opened 2025-04-16
feat: Add usdt/ethereum-lumia

ltyu:
### Description
This PR adds USDT/ethereum-lumia for Lumia Prism

### Backward compatibility
Yes


deployments/warp_routes/USDT/ethereum-lumia-deploy.yaml
```diff
@@ -0,0 +1,15 @@
+ethereum:
+  decimals: 6
+  mailbox: "0xc005dc82818d67AF737725bD4bf75435d065D239"
```
Xaroz, line 3: seems that deploy app still spit out mailbox but can we remove from here?

Xaroz: seems that deploy-app still spit out decimals and mailbox, but should we remove?

ltyu: yes good catch

Xaroz (approved):

PR #789, opened 2025-04-25
feat: add milkyway

paulbalaji:
### Description

feat: add milkyway
- grabbed chain metadata from cosmos registry https://github.com/cosmos/chain-registry/blob/master/milkyway/chain.json
- enroll on BSC

### Testing

manual


chains/milkyway/metadata.yaml
```diff
@@ -0,0 +1,38 @@
...
+domainId: 1296649291
+gasPrice:
+  amount: "0.03"
+  denom: umilk
+grpcUrls:
+  - http: https://grpc.mainnet.milkyway.zone/
+index:
+  chunk: 10
+  from: 2213100
+name: milkyway
```
yjamin, line 24: Lets use the live height of the mainnet: `2454618`

chains/milkyway/metadata.yaml
```diff
@@ -0,0 +1,38 @@
...
+deployer:
+  name: Abacus Works
+  url: https://www.hyperlane.xyz
+displayName: MilkyWay
+# Generated from: console.log(parseInt('0x'+Buffer.from('MILK').toString('hex')))
+domainId: 1296649291
+gasPrice:
+  amount: "0.03"
+  denom: umilk
+grpcUrls:
```
yjamin, line 19: I think gasPrice of 0.02 should be enought, double check with milkyway team tho. 

This is from the keplr chain registry:
https://github.com/chainapsis/keplr-chain-registry/blob/4098b94f68845dc0c3b2277c1fc91c3a649ed3a0/cosmos/milkyway.json#L120-L124

chains/milkyway/metadata.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - http: https://grpc.mainnet.milkyway.zone/
+index:
+  chunk: 10
+  from: 2213100
```
paulbalaji, line 23:
```suggestion
  from: 2454618
```

chains/milkyway/metadata.yaml
paulbalaji: Huh interesting, we have logic in infra that sets the value in the agent config by reading the cosmos registry

yjamin (approved):

chains/milkyway/metadata.yaml
yjamin: Ah okay, we can leave it as it is then 👍

chains/bsc/addresses.yaml
```diff
@@ -5,7 +5,7 @@ fallbackRoutingHook: "0x237E81f87F57Badad9e09f13CC676D986cA852e7"
 interchainAccountIsm: "0x9e22945bE593946618383B108CC5bce09eBA4C26"
 interchainAccountRouter: "0x32A07c1B7a7fe8D4A0e44B0181873aB9d64C16c1"
 interchainGasPaymaster: "0x78E25e7f84416e69b9339B0A6336EB6EFfF6b451"
-interchainSecurityModule: "0x006654622C311436f41C15cB0f236e7e8f2852d0"
+interchainSecurityModule: "0xCc99f457F00D5d7c4284045dA3D9Fb89635A952C"
```
paulbalaji, line 8:
```suggestion
interchainSecurityModule: "0x006654622C311436f41C15cB0f236e7e8f2852d0"
```

PR #795, opened 2025-04-27
feat: replace bad rpcs

xeno097:
### Description

Removes bad rpcs from the registry, re-enables the health tests and skips the check for flaky or dead rpcs

### Backward compatibility

- YES

### Testing

- YES

### DIFF
--- a/chains/berabartio/metadata.yaml
+++ b/chains/berabartio/metadata.yaml
@@ -23,6 +23,4 @@ nativeToken:
   symbol: BERA
 protocol: ethereum
 rpcUrls:
-  - http: https://bartio.rpc.berachain.com/
-  - http: https://bartio.drpc.org
-  - http: https://bera-testnet.nodeinfra.com
+  - http: https://berachain-testnet-rpc.publicnode.com
--- a/chains/berabartio/metadata.yaml
+++ b/chains/berabartio/metadata.yaml
@@ -23,4 +23,4 @@ nativeToken:
   symbol: BERA
 protocol: ethereum
 rpcUrls:
-  - http: https://berachain-testnet-rpc.publicnode.com
+  - http: https://bartio.drpc.org
--- a/test/health/rpc-health.test.ts
+++ b/test/health/rpc-health.test.ts
@@ -32,6 +32,7 @@ const CHAINS_TO_SKIP = new Set([
   'sonictestnet',
   // Flaky RPC
   'infinityvm',
+  'humanitytestnet',
 ]);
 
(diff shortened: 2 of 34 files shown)


chains/berabartio/metadata.yaml
```diff
@@ -23,6 +23,4 @@ nativeToken:
   symbol: BERA
 protocol: ethereum
 rpcUrls:
-  - http: https://bartio.rpc.berachain.com/
-  - http: https://bartio.drpc.org
-  - http: https://bera-testnet.nodeinfra.com
+  - http: https://berachain-testnet-rpc.publicnode.com
```
paulbalaji, line 26: this is wrong i think - because the bera bartio testnet has been deprecated?

paulbalaji: yeah the chainids are different
```
cast chain-id --rpc-url https://berachain-testnet-rpc.publicnode.com
80069
```

test/health/rpc-health.test.ts
```diff
@@ -15,7 +15,24 @@ import { chainAddresses, chainMetadata } from '../../dist/index.js';
 import { Mailbox__factory } from '@hyperlane-xyz/core';
 
 const CHAINS_TO_SKIP = new Set([
-  'worldchain', // only allows private RPC access
+  // only allows private RPC access
+  'worldchain',
```
paulbalaji, line 19: i think we can use a public worldchain endpoint now?

chains/berabartio/metadata.yaml
paulbalaji: do we have a test to ensure that every rpc added matches the chainid in the metadata?

test/health/rpc-health.test.ts
xeno097: I'll try but i suspect we will get rate limited

chains/berabartio/metadata.yaml
xeno097: damn i fell into the trap, many testnets seem to have been deprecated (probably a job for another pr)

xeno097: doesn't seems so otherwise it would have complained, seems a good idea to add it here now

xeno097: added the chain id check in the test

test/health/rpc-health.test.ts
xeno097: ok it worked fine it seems

paulbalaji (approved):

ltyu (approved):

PR #801, opened 2025-04-28
chore: Update remoteRouter names to domainIds

ltyu:
### Description
This PR Updates remoteRouter names to domainIds. Related https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6066

### Backward compatibility
Yes

### Testing
yarn check-warp-deploy

### DIFF
--- a/.changeset/tidy-donuts-train.md
+++ b/.changeset/tidy-donuts-train.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Update remoteRouter names to domainIds for zeronetwork USDC & USDT, and teasure SMOL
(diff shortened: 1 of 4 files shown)


paulbalaji (commented):
why does this only affect 3 files? 

doesn't it become more difficult to read if we define domainids instead of chainnames?

xeno097:
> why does this only affect 3 files?
> 
> doesn't it become more difficult to read if we define domainids instead of chainnames?

At least on the CLI side, we assume that the routers mapping has the domain IDs instead of the chain names

ltyu:
@paulbalaji 

> why does this only affect 3 files?
> 
> doesn't it become more difficult to read if we define domainids instead of chainnames?

We assume fully connected chains in the Checker when it is **not** defined. for the very few (the 3 that you see here) that are defined, the Check will validate all the connections (partial connections).

for now, our schema use domainId instead of name. We can change this in the future

paulbalaji:
> @paulbalaji
> 
> > why does this only affect 3 files?
> > doesn't it become more difficult to read if we define domainids instead of chainnames?
> 
> We assume fully connected chains in the Checker when it is **not** defined. for the very few (the 3 that you see here) that are defined, the Check will validate all the connections (partial connections).
> 
> for now, our schema use domainId instead of name. We can change this in the future

ahhh that makes sense, thanks for explaining

PR #806, opened 2025-04-30
feat: download full zip archives from GitHub

antigremlin:
### Description

This PR introduces downloading zip archives of the repository instead of fetching files one by one.

- If the authToken is set, it uses the REST archive download API.
- Otherwise, it uses the public source code archive API, trying the branch, tag, and commit-SHA URLs in turn.

### Backward compatibility

It should be backward compatible with the previous approach, but there is a discrepancy in the logic:

- We still use the REST API to get a list of files. This call could go through a proxy if the rate limit is exhausted. However, the actual files are read from the archive.

### Related issues

Part of ENG-1397.

### Testing

- The new logic is covered by unit tests.
- The main test is to run `yarn test:ci` for @hyperlane-xyz/infra.


src/registry/GithubRegistry.ts
```diff
@@ -75,6 +89,10 @@ export class GithubRegistry extends BaseRegistry implements IRegistry {
   public readonly repoName: string;
   public readonly proxyUrl: string | undefined;
   private readonly authToken: string | undefined;
+  // In-memory cache of archive file entries (relative path -> content as ArrayBuffer)
+  private archiveEntries?: Map<string, ArrayBuffer>;
+  // Promise tracking an in-flight archive download/unpack to dedupe parallel calls
+  private archiveEntriesPromise?: Promise<void>;
```
xeno097, line 95: 🔥

xeno097 (approved):

PR #825, opened 2025-05-02
chore: revert oUSDT extension and update lumia route

Xaroz:
### Description

- Revert oUSDT extension, major version because we are removing configs
- Update connections for lumia route, only allowing ethereum <> lumiaprism bridging for ETH

### Backward compatibility

No


nambrot (approved):

paulbalaji:
Please fix the monorepo CI because the revet would change back the route ID which may break monorepo warp configs test

PR #838, opened 2025-05-07
feat: check warp deploy in CI

paulbalaji:
### Description

feat: check warp deploy in CI
- new workflow to check warp deploy in CI
- gets the diff of the current PR and only checks the affected routes
- comment with the check-deploy result on PRs 

### Testing

copied seda route into this branch just for testing, will remove before merge


paulbalaji:
note: will remove the SEDA commit before merging. only included to show how the output is if there is a warp route in the PR.

PR #839, opened 2025-05-08
feat: add svmbnb

tkporter:
### Description

- Adds svmbnb metadata & addresses


chains/svmbnb/metadata.yaml
```diff
@@ -0,0 +1,26 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://explorer.svmbnbmainnet.soo.network/
+    family: other
+    name: svmBNB Explorer
+    url: https://explorer.svmbnbmainnet.soo.network/
```
paulbalaji, line 6: nit but i like to omit trailing slashes

chains/svmbnb/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+blocks:
+  confirmations: 1
+  estimateBlockTime: 0.4
+  reorgPeriod: 0
+chainId: 574456
+deployer:
+  name: Abacus Works
+  url: https://www.hyperlane.xyz
+displayName: svmBNB
+domainId: 574456
```
paulbalaji, line 16: how was this derived?
```suggestion
domainId: 574456
```

tkporter: Because it's svm it's arbitrary, but chatgpt says `5744` is most close to "SVM" in leetspeak, and 56 is the domain ID of bsc. I checked to make sure there are no conflicts out there

paulbalaji: gotcha, asking because weve been using this for cosmos and was curious is there was a similar algorithm for sealevel https://github.com/hyperlane-xyz/hyperlane-registry/blob/2df84ec76c2637e6e3fd9473d921c9563dd46ee1/chains/milkyway/metadata.yaml#L14

paulbalaji: no biggie for this PR though!

paulbalaji (approved):

chains/svmbnb/metadata.yaml
tkporter: ah gotcha

PR #854, opened 2025-05-15
feat: add starknetsepolia + paradexsepolia metadata

paulbalaji:
### Description

feat: add starknetsepolia + paradexsepolia metadata


chains/paradexsepolia/metadata.yaml
```diff
@@ -0,0 +1,26 @@
+# yaml-language-server: $schema=../schema.json
+blockExplorers:
+  - apiUrl: https://voyager.testnet.paradex.trade/api
+    family: other
+    name: Paradex testnet
+    url: https://voyager.testnet.paradex.trade
+blocks:
+  confirmations: 0
+  estimateBlockTime: 4
+  reorgPeriod: 0
```
paulbalaji, line 10: @aroralanuk why are confs/reorg both 0 for starknet

yjamin: Yeah that sounds weird to me as well. But the agents apparently work with that config

PR #858, opened 2025-05-16
chore: rename oUSDT warp route path to match token symbol defined in config

Mo-Hussain:
### Description

- oUSDT file path and therefore the derived warp route id does not match the token symbol defined in the config, this is leading to the dashboards going out of sync
- This change will also need a monorepo PR


.changeset/tall-falcons-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': minor
+---
+
+rename path to oUSDT route so that the symbol is consistant, USDT -> oUSDT
```
paulbalaji, line 5:
```suggestion
rename path to oUSDT route so that the symbol is consistent, USDT -> oUSDT
```

antigremlin (approved):

Mo-Hussain:
> if we are moving this can we also move the `oUSDT-logo.svg` file to the `oUSDT` folder and then change the `logoURI` path in the `config.yaml`?

Do we need anymore changes in the warp ui for this?

Xaroz:
> > if we are moving this can we also move the `oUSDT-logo.svg` file to the `oUSDT` folder and then change the `logoURI` path in the `config.yaml`?
> 
> Do we need anymore changes in the warp ui for this?

nope, no changes are required from the warp UI side since the resulting `warpRouteId` would still be the same

PR #875, opened 2025-05-21
Add new set of ICAs with call commitment support

nambrot:
### Description

New ICA routers with call commitments


paulbalaji (commented):
This will break existing ICAs on those chains

PR #885, opened 2025-05-23
fix: only download zip archives when in Node.js

antigremlin:
### Description

This PR fixes the zip-downloading issue experienced by browser-based registry clients. We tried to download the zip archives in a way that's prevented by CORS. There's no quick workaround, so the code now reverts to old behaviour when running in a browser.

### Backward compatibility

Yes

### Testing

- No new unit-tests added
- Manually tested from hyperlane-explorer


paulbalaji (commented):
patch changeset as well please

paulbalaji (approved):

PR #891, opened 2025-05-26
refactor: simplify warp configs

christopherbrumm:
### Description

This PR simplifies warp route configs by removing
- redundant owners
- default ISM configuration.

To achieve this, the following script has been used: https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/64ae5840aa10552194ddb9aab2875f6aa042a738

The script can be used to check what has been removed and review the correct behaviour. It might make sense to create a workflow based on this script to prevent redundant configs.

### Backward compatibility

Yes

### Testing

N/A


ltyu:
Can you make sure those check-warp-deploy are green, unless there is a valid reason they aren't

PR #903, opened 2025-05-27
chore: update GithubRegistry - getWarpRoutes

Xaroz:
### Description

- Update GithubRegistry getWarpRoutes to fetch combined warpRouteConfigs.yaml file instead

### Backward compatibility

Yes


src/registry/BaseRegistry.ts
```diff
@@ -37,6 +37,8 @@ export abstract class BaseRegistry implements IRegistry {
   protected isMetadataCacheFull: boolean = false;
   protected addressCache?: ChainMap<ChainAddresses>;
   protected isAddressCacheFull: boolean = false;
+  protected warpRouteCache?: WarpRouteConfigMap;
```
paulbalaji, line 40: should there be some TTL on this?

PR #904, opened 2025-05-28
fix: set correct reorgPeriod & blocktime

yjamin:
### Description
Blocktime of Starknet sepolia is 30s, not 5. Also the agents often see errors of missing blocks, that is a result of the 0 reorgPeriod. Set that to 1 to have some buffer. That missing buffer also results in the scraper wrongfully not indexing TXs.


paulbalaji (commented):
add isms at 0 addr too

chains/paradexsepolia/metadata.yaml
```diff
@@ -7,7 +7,7 @@ blockExplorers:
 blocks:
   confirmations: 1
   estimateBlockTime: 4
```
paulbalaji, line 9: Paradex is still 4?

yjamin: Let me check, the explorer shows a block time of ~6minutes

paulbalaji: are we sure this is the _actual_ block time, or just the estimated block time with on-demand block production

yjamin: Take a look for yourself: https://voyager.testnet.paradex.trade/blocks

yjamin: However the blocktime seems to be flaky. I would just set it to 30sec as well. This should be high enough and won't cause any problems. At lest from the agents perspective. Is this fine?

.changeset/heavy-moles-breathe.md
paulbalaji, line 1: don't really need changesets for unit tests but hey no harm!

paulbalaji (approved):

PR #912, opened 2025-05-30
fix(metadata): switch from STRK to ETH

aroralanuk:
### Description

- the protocol fee token is using STRK but we've been use ETH in the relayer operations so far (should upgrade soon)


paulbalaji (approved):
pls add patch changeset

PR #917, opened 2025-06-02
chore: Add mailbox to solana mainnet + testnet metadata

MaanavKhaitan:
### Description

- Adds `mailbox` field to the chain metadata for `solanatestnet` and `solanamainnet`
- This is required when using the hyperlane CLI to send tokens over a warp route from an SVM chain, because the CLI requires the `mailbox` field for sealevel tokens (https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/823234337e2b3f3c9b6e19c026303610e4729040/typescript/sdk/src/token/Token.ts#L246)

### Backward compatibility

Yes

### DIFF
--- a/chains/solanamainnet/metadata.yaml
+++ b/chains/solanamainnet/metadata.yaml
@@ -15,6 +15,7 @@ deployer:
 displayName: Solana
 domainId: 1399811149
 gasCurrencyCoinGeckoId: solana
+mailbox: E588QtVUvresuXq2KoNEwAmoifCzYGpRBdHByN9KQMbi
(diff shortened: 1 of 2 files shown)


paulbalaji:
there is an addresses yaml already. The cli doesn't support sending to sealevel chains. To test SVM<>EVM bridging I recommend spinning up an instance of the warp-ui-template instead

MaanavKhaitan:
I am testing sending from sealevel chains, not to sealevel chains. I got a transfer from SVM --> EVM to work in the CLI by using `hyperlane warp send`

PR #918, opened 2025-06-02
feat(starknet): updated `starknet` domainId and redeployment of core contracts

aroralanuk:
### Description

- TLDR: scraper db doesn't support domainIds beyond int32 max bc the db's id is set as signed integer. So I redeployed mailbox and contracts dependent on the mailbox (hooks and validator_announce)

### Backward compatibility

Yes

### Testing

Inspecting on the explorer

### DIFF
--- a/test/unit/chains.test.ts
+++ b/test/unit/chains.test.ts
@@ -11,7 +11,12 @@ import { isAbacusWorksChain } from '../../src/utils.js';
 import { expect } from 'chai';
 
 describe('Chain metadata', () => {
+  const skippedChainsList = ['inclusivelayertestnet'];
   for (const [chain, metadata] of Object.entries(chainMetadata)) {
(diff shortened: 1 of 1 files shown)


chains/starknet/metadata.yaml
```diff
@@ -13,8 +13,8 @@ deployer:
   name: Abacus Works
   url: https://www.hyperlane.xyz
 displayName: Starknet
-# keccak256("starknet") % MAX_UINT32
-domainId: 3047739480
+# keccak256("starknet") % MAX_INT32
```
paulbalaji, line 16: confirmed with
```sh
cast keccak "starknet" | cast to-dec | xargs -I{} echo "{} % (2^31 - 1)" | bc
```

paulbalaji (approved):

PR #927, opened 2025-06-03
chore: don't check deleted routes

paulbalaji:
### Description

chore: don't check deleted routes
- currently the diffing just looks at anything that changed, so if a file was deleted then we would try to check-warp-deploy that. but it does not exist. e.g 
image
- by adding `--diff-filter=ARM`, we're only considering additions/renames/modifications

### Testing

- tested by remaking the ousdt PR but with this change cherry-picked: https://github.com/hyperlane-xyz/hyperlane-registry/pull/928
- correctly didn't include removed routes:
image
- correctly [reported](https://github.com/hyperlane-xyz/hyperlane-registry/pull/928#issuecomment-2934887322) staging ✅ and prod ❌:
image

### DIFF
--- a/scripts/check-warp-deploy.sh
+++ b/scripts/check-warp-deploy.sh
@@ -14,7 +14,8 @@ export BASE_COMMIT="$1"
 export HEAD_COMMIT="$2"
 
 WARP_ROUTE_IDS=$(
-    git diff "$BASE_COMMIT".."$HEAD_COMMIT" --name-only |
+    # ARM = Additions, Renames, Modifications
+    git diff --diff-filter=ARM "$BASE_COMMIT".."$HEAD_COMMIT" --name-only |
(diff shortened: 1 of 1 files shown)


scripts/check-warp-deploy.sh
```diff
@@ -14,7 +14,8 @@ export BASE_COMMIT="$1"
 export HEAD_COMMIT="$2"
 
 WARP_ROUTE_IDS=$(
-    git diff "$BASE_COMMIT".."$HEAD_COMMIT" --name-only |
+    # ARM = Additions, Renames, Modifications
+    git diff --diff-filter=ARM "$BASE_COMMIT".."$HEAD_COMMIT" --name-only |
```
xeno097, line 18: nearly `ASMR`

xeno097 (approved):

PR #931, opened 2025-06-04
feat: deploy to botanix, katana

paulbalaji:
### Description

feat: deploy to botanix, katana


chains/arcadia/metadata.yaml
```diff
@@ -14,7 +14,8 @@ deployer:
   url: https://www.hyperlane.xyz
 displayName: Arcadia
 domainId: 4278608
-gasCurrencyCoinGeckoId: khalani-aip # TODO: update once we have a valid id
+# TODO: update to khalani-aip once we have a valid id
+gasCurrencyCoinGeckoId: ethereum
```
tkporter, line 18: curious why ethereum? is AIP close to ETH in price?

chains/botanix/metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+nativeToken:
+  decimals: 18
+  name: Bitcoin
+  symbol: BTC
+protocol: ethereum
+rpcUrls:
+  - http: https://botanixrpc.com
+technicalStack: other
+transactionOverrides:
+  maxFeePerGas: 1000007
```
tkporter, line 28: do you expect this to change much? as to my knowledge we've tended not to put tx overrides in the registry and have instead put them in infra, but happy either way if you think this won't cause problems. Also as we move further from infra tooling and to CLI this probably makes more sense

tkporter (approved):

chains/arcadia/metadata.yaml
paulbalaji: in the tokenPrices.json we've always been using the eth price - this just reflects that directly in registry. there's some aip/eth redemption mechanism iirc which is why we chose eth

chains/botanix/metadata.yaml
paulbalaji: will likely need to be tweaked later, still a new chain so some rough riding for users. but without this change, anyone will pay 1000x more than required. fun fact we spent more gas on 2txs using the default chain's gas price than the entire core deploy itself with the capped fee

PR #948, opened 2025-06-11
feat: deploy SOL warp route between solanamainnet and starknet

aroralanuk:
### Description

- deployed warp route between solanamainnet <> starknet

### DIFF
--- a/deployments/warp_routes/SOL/solanamainnet-starknet-config.yaml
+++ b/deployments/warp_routes/SOL/solanamainnet-starknet-config.yaml
@@ -1,0 +1,20 @@
+# yaml-language-server: $schema=../schema.json
+tokens:
+  - addressOrDenom: EfSiyLayPpfZ7uGetjmjxYh6r3LjoJjmHkkxRJmDpgaT
+    chainName: solanamainnet
+    connections:
+      - token: starknet|starknet|0x01e70aedffd376afe33cebdf51ed5365131dccb2a5b2cb36d02b785442912b9b
+    decimals: 9
+    logoURI: /deployments/warp_routes/SOL/logo.svg
+    name: Solana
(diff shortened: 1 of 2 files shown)


deployments/warp_routes/SOL/solanamainnet-starknet-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+tokens:
+  - addressOrDenom: EfSiyLayPpfZ7uGetjmjxYh6r3LjoJjmHkkxRJmDpgaT
+    chainName: solanamainnet
+    connections:
+      - token: starknet|starknet|0x01e70aedffd376afe33cebdf51ed5365131dccb2a5b2cb36d02b785442912b9b
+    decimals: 9
+    logoURI: /deployments/warp_routes/SOL/logo.svg
+    name: Solana
+    standard: SealevelHypNative
+    symbol: SOL
```
paulbalaji, line 11: do we need the gecko id here to track collateral?

deployments/warp_routes/warpRouteConfigs.yaml
```diff
@@ -5957,6 +5978,108 @@ WETH/artela-base:
       name: Wrapped Ether
       standard: EvmHypCollateral
       symbol: WETH
+WFRAGJTO/fragmetric:
```
tkporter, line 5981: this is resulting from a `yarn && yarn clean && yarn build`. I think wfragJTO was renamed to WFRAGJTO but for some reason this hasn't been reflected in the built warpRouteConfigs.yaml?

paulbalaji: the wfragJTO version is the correct one

image

PR #953, opened 2025-06-12
fix: native denom padding on paradex

yjamin:
### Description
Add padding to paradex denom to fill 32 bytes


tkporter (approved):
can you add a changeset

paulbalaji (approved):

PR #962, opened 2025-06-13
feat: redeploy on rome testnet

paulbalaji:
### Description

feat: redeploy on rome testnet


chains/rometestnet/metadata.yaml
```diff
@@ -9,9 +9,6 @@ blocks:
   estimateBlockTime: 4
   reorgPeriod: 1
 chainId: 200018
-deployer:
```
tkporter, line 12: is this also used as a rough metric for a "live" chain?

tkporter: also kinda curious if we should just rm the old `rometestnet` because it's gone

tkporter (approved):

PR #965, opened 2025-06-16
chore: add coingecko id for monitor

cmcewen:


xeno097 (approved):
the coingecko id can be omitted for synthetic chains

ltyu (commented):
I believe we only use gecko id for collateral: https://www.notion.so/hyperlanexyz/EVM-Only-Route-1216d35200d68032b34bf8cbc6ce9555?source=copy_link#1696d35200d68085a1e0f38cdb2a96dd

PR #986, opened 2025-06-23
chore: add oUSDT coingecko Id

mbreithecker:
### Backward compatibility

Yes

### DIFF
--- a/deployments/warp_routes/oUSDT/production-config.yaml
+++ b/deployments/warp_routes/oUSDT/production-config.yaml
@@ -2,6 +2,7 @@
 tokens:
   - addressOrDenom: "0x4F0654395d621De4d1101c0F98C1Dba73ca0a61f"
     chainName: base
+    coinGeckoId: openusdt
     collateralAddressOrDenom: "0x1217BfE6c773EEC6cc4A38b5Dc45B92292B6E189"
     connections:
       - token: ethereum|bob|0x155332Fad14bdE126255E6Ab6A09Fe88D2864294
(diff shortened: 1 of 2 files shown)


deployments/warp_routes/oUSDT/production-config.yaml
paulbalaji, line 1: isn't the gecko id only needed on collateral chains, which we already do?

https://github.com/hyperlane-xyz/hyperlane-registry/blob/4ac34c041ab34f1603bf7334bda2e79b04ead187/deployments/warp_routes/oUSDT/production-config.yaml#L80
https://github.com/hyperlane-xyz/hyperlane-registry/blob/4ac34c041ab34f1603bf7334bda2e79b04ead187/deployments/warp_routes/oUSDT/production-config.yaml#L57

paulbalaji: interesting - in most cases we keep the collateral/token name the same but for ousdt it's intentional that you'd have a different collateral name to the token

i feel for accurate metrics here we should be using `tether` as that's the underlying token, but if you need `ousdt` then that's fine with me too. you'll only need to update `ethereum` and `celo` for this though

mbreithecker: I have no preference here. So right now for analytics we use the data in the configs to map a coingecko id to a symbol. I think openusdt and usdt should almost always be the same value. 

Is it safe to use the folder name for the symbol and then the first coingecko-id I can find in the files?
Are there some examples where deployments have different prices on source and destination chains?

paulbalaji: ok lets stick with `ousdt` then

PR #999, opened 2025-06-26
chore: update to new ism structure

yjamin:
### Description
Update paradex & starknet default ISM addresses to new ISM structure.


paulbalaji (commented):
paradex is already on the new setup?

PR #1059, opened 2025-07-18
chore: Fix route typo

ltyu:
### Description
Fixes typo in get warp cores route

### DIFF
--- a/.changeset/odd-numbers-add.md
+++ b/.changeset/odd-numbers-add.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Fix typo for HttpClientRegistry route /warp-route/core


src/registry/HttpClientRegistry.ts
```diff
@@ -134,7 +134,7 @@ export class HttpClientRegistry implements IRegistry {
     if (filter?.label) {
       queryParams.set('label', filter.label);
     }
-    return this.fetchJson<WarpRouteConfigMap>(`/warp-routes/core?${queryParams.toString()}`);
+    return this.fetchJson<WarpRouteConfigMap>(`/warp-route/core?${queryParams.toString()}`);
```
paulbalaji, line 137: in the repo it's under `deployments/warp_routes` tho

is this an expected divergence between the repo and the api?

paulbalaji: ah yes i see other places use `/warp-route` https://github.com/hyperlane-xyz/hyperlane-registry/blob/5cc34e1e7d2c456e770d36f95d7caa6eeccc507e/src/registry/HttpClientRegistry.ts#L126

PR #1073, opened 2025-07-31
feat: deploy camp

Mo-Hussain:
### Description

- deploy camp


mbreithecker:
Looks good to me. Should we add the index: from property? There have already been some blocks produced on their network

PR #1081, opened 2025-08-06
chore: deprecate chains

paulbalaji:
### Description

Deprecate chains bouncebit, conflux, conwai, deepbrainchain, duckchain, evmos, flame, nero, rivalz, rootstockmainnet, sanko, telos, unitzero, xpla, zklink


chains/conwai/metadata.yaml
```diff
@@ -9,9 +12,6 @@ blocks:
   estimateBlockTime: 2
   reorgPeriod: 0
 chainId: 668668
-deployer:
```
Mo-Hussain, line 12: Why do we remove the deployer info?

paulbalaji: good q tbh, I think it's because originally we used a deployer==AW check to determine which chains to include in the docs for like isms/mailbox/domains etc

https://github.com/hyperlane-xyz/v4-docs/blob/f214eb1706585cafbcbdd647a94e0fc2f81a1ba4/scripts/registry.js#L11-L14

this "availability" property also hasn't been used properly til recently, so we haven't updated these scripts downstream to account for the property

paulbalaji: I'll ask eda to update the docs generators

i dont think it's strictly necessary to remove the deployer metadata but if it's not actively supported idk if it makes sense to keep it either

paulbalaji: oh actually the real reason for removing the `deployer` is because we've also removed the `addresses`

Mo-Hussain (approved):

PR #1112, opened 2025-08-27
feat: radix deployment

yjamin:


chains/addresses.yaml
paulbalaji, line 1: reminder for future: you don't need to commit the combined addresses.yaml/metadata.yaml

paulbalaji (approved):

chains/addresses.yaml
yjamin: Oh yeah I just built it because i was already linking my monorepo and it didn't find the new chain

PR #1126, opened 2025-09-01
chore: add `HOLO` metadata

christopherbrumm:
### Description

chore: add `HOLO` metadata


paulbalaji:
`yarn prettier`

christopherbrumm:
> `yarn prettier`

Executed `prettier`, then executed `lint`. Can we change the configs so the jobs don't overwrite each others changes?

PR #1147, opened 2025-09-15
chore: update pulsechain block explorer

Xaroz:
### Description

Update `pulsechain` block explorer `apiUrl` and `url`
### Backward compatibility

Yes


chains/pulsechain/metadata.yaml
```diff
@@ -1,13 +1,13 @@
...
+  - apiUrl: https://scan.pulsechain.box/api
     family: blockscout
     name: Pulsechain Explorer
-    url: https://scan.pulsechainfoundation.org/#/
+    url: https://scan.pulsechain.box
   - apiUrl: https://api.scan.pulsechain.com/api
     family: blockscout
     name: Pulsechain Explorer
-    url: https://oldscan.gopulse.com/#/
+    url: https://scan.pulsechainfoundation.org/#/
```
xeno097, line 10: I think you should only change the block epxlorer info for `https://oldscan.gopulse.com/#/` and leave the others as they are as each block explorer has a different api

Xaroz: The diff looks a bit weird but that is essentially what it is, `oldscan` pulse was replaced with `scan.pulsechain.box` and `scan.pulsechainfoundation.org` was moved down as second item in the index per request

xeno097 (approved):

PR #1174, opened 2025-10-02
chore: fix SOL/radix-solanamainnet filename

mbreithecker:


mbreithecker:
> what's wrong with the file name? the convention is not to need the full list of chains anymore?

Ohh didnt know that. Thought I approved the previous PR to early and later saw that the pattern is different. 
Is it okay to leave it then?

paulbalaji:
> Ohh didnt know that. Thought I approved the previous PR to early and later saw that the pattern is different. Is it okay to leave it then?

yup! we made the change around may/june to support stable warp IDs 
- https://www.notion.so/hyperlanexyz/Stable-Warp-Route-IDs-1d76d35200d680f39c7ff2060b3ca9fe?source=copy_link

PR #1179, opened 2025-10-06
chore: update galactica coingecko IDs

paulbalaji:
### Description

chore: update galactica coingecko IDs


paulbalaji:
can ignore the check-warp-deploy error as it's the usual issue on the CI job with hyp native

PR #1474, opened 2026-04-13
fix(chains): fix hyperevm blockexplorer api url + add monadscan explorer info

xeno097:
### Description

Remove v2 suffix from hyperevm block explorer api + add monadscan info

### Backward compatibility

- yes

### Testing

- Cli locally

### DIFF
--- a/chains/hyperevm/metadata.yaml
+++ b/chains/hyperevm/metadata.yaml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=../schema.json
 blockExplorers:
-  - apiUrl: https://www.hyperscan.com/api/
+  - apiUrl: https://www.hyperscan.com/api
     family: blockscout
     name: HyperEVM Explorer
     url: https://www.hyperscan.com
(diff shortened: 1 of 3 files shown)


chains/hyperevm/metadata.yaml
```diff
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=../schema.json
 blockExplorers:
-  - apiUrl: https://www.hyperscan.com/api/v2
+  - apiUrl: https://www.hyperscan.com/api
```
paulbalaji, line 3: how come we dont need v2 anymore https://github.com/hyperlane-xyz/hyperlane-registry/pull/1344/changes

chains/monad/metadata.yaml
```diff
@@ -1,5 +1,10 @@
 # yaml-language-server: $schema=../schema.json
 blockExplorers:
+  - apiKey: GP69JEAP2W7YFJT9ZJTEPGQT6Y6KMW44ZN
+    apiUrl: https://api.etherscan.io/v2/api?chainid=143
+    family: etherscan
```
paulbalaji, line 5: can etherscan just die 😭 i guess theres no blockscout for monad

paulbalaji (approved):

chains/hyperevm/metadata.yaml
xeno097: `/v2` made api requests fail even if the contracts were verified on the explorer ahaha

PR #1484, opened 2026-04-20
chore: validate scale consistency between warp route deploy and config

Xaroz:
## Summary
- Adds a check in `scripts/validate-file-data.js` requiring each chain with a `scale` in `*-deploy.yaml` to have a matching `scale` on the corresponding token in `*-config.yaml`.
- Runs as part of the existing `validate-file-data` step in CI.

This PR intentionally only contains the validator change so CI surfaces any existing mismatches. Config fixes will follow in a separate commit.

## Test plan
- [ ] CI `test` job runs `validate-file-data.js` and reports any scale mismatches
- [ ] Follow-up commit fixes mismatches and re-runs CI to green

### DIFF
--- a/scripts/validate-file-data.js
+++ b/scripts/validate-file-data.js
@@ -12,6 +12,7 @@ const noLogoFileError = [];
 // warp route warnings / errors
 const noConfigFileWarning = [];
 const invalidLogoURIPathError = [];
+const scaleMismatchError = [];
 
 function validateChains() {
   if (!fs.existsSync(chainsDir)) {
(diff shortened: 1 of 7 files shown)


paulbalaji:
Could this live in `test/unit/warp-routes.test.ts` instead? That file already iterates every warp route via `FileSystemRegistry` and has a parallel `describe('Warp Deploy Configs', ...)` block using `getWarpDeployConfigs()`, so the wiring is basically there:

```ts
describe('Warp scale consistency', () => {
  const reg = new FileSystemRegistry({ uri: BASE_URI });
  const cores = reg.getWarpRoutes();
  const deploys = reg.getWarpDeployConfigs();

  for (const id of Object.keys(deploys)) {
    it(`${id} deploy scale matches config scale per chain`, () => {
      const deploy = deploys[id];
      const core = cores[id];
      if (!core) return;
      for (const [chain, cfg] of Object.entries(deploy)) {
        if (!('scale' in cfg)) continue;
        const token = core.tokens.find((t) => t.chainName === chain);
        expect(normalizeScale(token?.scale)).to.deep.equal(normalizeScale(cfg.scale));
      }
    });
  }
});
```

Tradeoffs vs keeping it in `validate-file-data.js`:
- **Unit test**: per-route failure granularity in CI output, reuses `FileSystemRegistry` loading (no re-reading YAML by path), sits alongside the existing warp cross-file checks.
- **validate-file-data**: single aggregated error report, no build step needed.

Since the other warp cross-file checks (logoURI presence/consistency) already live in the unit tests, putting this there too seems more consistent — wdyt?

scripts/validate-file-data.js
```diff
@@ -42,6 +43,47 @@ function validateChains() {
...
+  if (scale == null) return null;
+  if (typeof scale === 'number') return { numerator: 1, denominator: scale };
+  if (typeof scale === 'object' && 'numerator' in scale && 'denominator' in scale) {
+    return { numerator: scale.numerator, denominator: scale.denominator };
+  }
+  return null;
+}
+
+function scaleEqual(a, b) {
+  if (a == null || b == null) return false;
```
xeno097: why not use is nullish here?

xeno097 (approved):

paulbalaji (approved):
Nicely done — delegating to the SDK's `normalizeScale` and keeping the null-for-absent wrapper is cleaner than the original handwritten version, and the dedicated unit tests for the wrapper make the semantics explicit. Sitting next to the existing warp cross-file checks feels like the right home.

PR #1491, opened 2026-04-22
fix: fluent naming

troykessler:


paulbalaji (approved):

paulbalaji:
image
idk why it's complaining but your branch is 10 commits behind main so maybe could be something idk?

PR #1522, opened 2026-05-12
feat(warp-routes): deployed aleotestnet routes

xeno097:
### Description

Adds the 
- sol
- wbtc
- usdc
- usdt

testnet warp routes to connect to aleo

### Backward compatibility

-yes

### Testing

local ui instance


chains/aleotestnet/metadata.yaml
```diff
@@ -24,4 +24,4 @@ nativeToken:
   symbol: ALEO
 protocol: aleo
 rpcUrls:
-  - http: https://api.explorer.provable.com/v1
+  - http: https://api.explorer.provable.com/v2
```
troykessler, line 27: isnt this the aleo mainnet rpc?

xeno097: noup, the base of the rpc url is the same for both mainnet and testnet what changes is that the sdk code appends `/testnet` or `/mainnet` based on the registry config

troykessler: ah I see

troykessler (approved):

PR #1534, opened 2026-05-18
fix: point WARD/ward-base at the production Base↔Warden route

mn13:
### Description

`WARD/ward-base` currently resolves to `0x6252BefAc2E1a8f7CbD3d43C6B8479790F37a98f` on both Base and Warden, which holds only ~595 WARD of bridged supply. The route that's actually in production use is a separate deployment with **18.5M WARD of bridged volume**:

- Base: `0xf09e4C8193F16019F0573F370F9A997b11f56638` (HypSynthetic, owned by the Base Safe `0x026AA469…`)
- Warden: `0xc5ADACe5E2250e5497CF1eF7B4f100e10099e4c4` (HypNative, owned by the Warden Safe `0x873f6699…`)

This PR updates the yaml so `WARD/ward-base` resolves to the production pair. Both routes use the same Mailbox (`0x15eECB28…`) and verify through the same 2-of-4 multisig (`0xE8935e…`), so the security model is unchanged. The original `0x6252…` deployment is left untouched on-chain.

### Backward compatibility

No. Tooling that previously dispatched through `WARD/ward-base` will now hit a different on-chain pair. Holders of synthetic WARD on the old Base `0x6252…` contract are not affected at the contract level — they can still bridge back by interacting with that contract directly — but the route ID will no longer resolve to it. Outstanding supply on that route is small (~595 WARD).

### Testing

Yes. End-to-end verified with a live Warden→Base transfer using the new addresses: native escrow on `0xc5ADACe5…` and synthetic mint on `0xf09e4C81…` both incremented as expected.

- Origin transaction (dispatch on Warden): https://explorer.wardenprotocol.org/tx/0xcdfcdfed5dd006743b5d6e51a1c975a9e6da15137ee31927950b64636132e0a3
- Destination transaction (mint on Base): https://basescan.org/tx/0xbabf53194317d56c8e99445c94659ffb5b8b87f1f5860ad85fdf46aa6cc2f8d7

### DIFF
--- a/.changeset/ward-base-canonical-addresses.md
+++ b/.changeset/ward-base-canonical-addresses.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+The WARD/ward-base route now points at the production Base↔Warden deployment (Base 0xf09e4C8193F16019F0573F370F9A997b11f56638 ↔ Warden 0xc5ADACe5E2250e5497CF1eF7B4f100e10099e4c4).
--- a/deployments/warp_routes/WARD/ward-base-config.yaml
+++ b/deployments/warp_routes/WARD/ward-base-config.yaml
@@ -1,18 +1,18 @@
 # yaml-language-server: $schema=../schema.json
 tokens:
-  - addressOrDenom: "0x6252BefAc2E1a8f7CbD3d43C6B8479790F37a98f"
+  - addressOrDenom: "0xf09e4C8193F16019F0573F370F9A997b11f56638"
     chainName: base
     connections:
-      - token: ethereum|wardenprotocol|0x6252BefAc2E1a8f7CbD3d43C6B8479790F37a98f
+      - token: ethereum|wardenprotocol|0xc5ADACe5E2250e5497CF1eF7B4f100e10099e4c4
     decimals: 18
     logoURI: /deployments/warp_routes/WARD/logo.svg
     name: WARD
     standard: EvmHypSynthetic
     symbol: WARD
-  - addressOrDenom: "0x6252BefAc2E1a8f7CbD3d43C6B8479790F37a98f"
+  - addressOrDenom: "0xc5ADACe5E2250e5497CF1eF7B4f100e10099e4c4"
     chainName: wardenprotocol
     connections:
-      - token: ethereum|base|0x6252BefAc2E1a8f7CbD3d43C6B8479790F37a98f
+      - token: ethereum|base|0xf09e4C8193F16019F0573F370F9A997b11f56638
     decimals: 18
     logoURI: /deployments/warp_routes/WARD/logo.svg
     name: WARD


paulbalaji (commented):
## On-chain validation — LGTM

I validated every concrete claim in the description against on-chain state. All check out. Recommend merge.

### Validation results

| Claim | Result |
|---|---|
| Base `0xf09e…` is HypSynthetic (WARD), owned by Base Safe `0x026AA469…` | ✅ owner matches; `wrappedToken()` reverts (synthetic); `symbol()="WARD"` |
| Warden `0xc5ADACe…` is HypNative, owned by Warden Safe `0x873f6699…` | ✅ owner matches; holds 18,528,747.200002 WARD as native balance |
| 18.5M WARD bridged volume on the new pair | ✅ Base `totalSupply` == Warden native balance == 18,528,747.200002 WARD (escrow perfectly matches minted) |
| ~595 WARD stranded on old `0x6252…` | ✅ 595.500007 WARD on both sides |
| Both new contracts share mailbox `0x15eECB28…` | ✅ confirmed |
| Routers mutually enrolled on new pair | ✅ Base→8765 router = `0xc5ADACe…`; Warden→8453 router = `0xf09e…` |
| Verifies through 2-of-4 multisig `0xE8935e…` | ✅ ISM tree on both chains terminates at `0xE8935e…` with validators `[0xB117…, 0xEc55…, 0x49CF…, 0x68c3…]`, threshold 2 |
| End-to-end test transactions | ✅ Warden dispatch tx hit `0xc5ADACe…`; Base receipt processed by mailbox `0x15eECB…` |

### Findings worth calling out

**1. ISM structure is slightly different (not a regression).**
- Old route Base ISM = AggregationISM(threshold 1-of-2) → one branch is a RoutingISM to the `0xE8935e` multisig, the other branch is an unidentified module (`0x36BF6934…`, reverted on every inspection signature I tried).
- New route Base ISM = AggregationISM(threshold 1-of-1) → single RoutingISM straight to the same `0xE8935e` multisig.
- The actual validator set signing messages is identical. The old route had an extra branch that could theoretically have provided an alternate verification path; the new route removes it. Arguably a tightening, not a loosening.

**2. Ownership change is intentional and within-team, but worth flagging.**
- Old route: both Base and Warden sides owned by Warden Safe (`0x873f6699…`).
- New route: Base side owned by Base Safe (`0x026AA469…`), Warden side still by Warden Safe.
- I checked both Safes — same 7 signers on each, but the **Base Safe is 4-of-7 and the Warden Safe is 3-of-7**. So the new Base-side admin policy is one signature stricter than the old admin (Warden Safe). FYI — no action needed, but worth noting that the two sides of the route now have asymmetric admin thresholds.

**3. Non-canonical mailbox (pre-existing, not introduced here).**
- This route uses Base mailbox `0x15eECB28…`, but `chains/base/addresses.yaml` lists the canonical Base mailbox as `0xeA87ae93Fa0019a82A727bfd3eBd1cFCa8f64f1D`. The `0x15eECB28…` mailbox is a legitimate Hyperlane mailbox on Base (correct `localDomain()=8453`, owned by the Base Safe) — just a separate deployment the Warden team runs. Same situation as the old route, so not introduced by this PR; flagging as an FYI for tooling that assumes the canonical mailbox.

### YAML diff itself
Correct, consistent with on-chain pair, changeset entry is appropriate.

paulbalaji (approved):
Approving — see prior review comment for full on-chain validation. All claims verified against state on Base and Warden.

PR #1542, opened 2026-05-26
feat: moonpay polygon extension

troykessler:


troykessler:
closed in favor of https://github.com/hyperlane-xyz/hyperlane-registry/pull/1551

PR #1548, opened 2026-06-04
fix: reduce katana reorgPeriod to 1

haggis-hyperlane:
## Summary
- Reduces `reorgPeriod` for Katana (chainId 747474) from `5` to `1` in `chains/katana/metadata.yaml`.

Requested by @Nam Chu Hoai.


paulbalaji (changes requested):
needs changeset

paulbalaji (approved):

PR #1551, opened 2026-06-05
feat: moonpay extension

troykessler:
### Description

Extended moonpay route to polygon, bsc and katana

Deploy configs generated with config getters: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8859


chains/katana/metadata.yaml
```diff
@@ -24,3 +24,6 @@ protocol: ethereum
 rpcUrls:
   - http: https://rpc.katana.network
 technicalStack: polygoncdk
+transactionOverrides:
+  maxFeePerGas: 10000000
+  maxPriorityFeePerGas: 10000000
```
paulbalaji, line 29: is this for all cli usage or just for agents? if just for agents we can add the overrides in the monorepo

paulbalaji (commented):
## Review summary

Nice work extending the routes to polygon/bsc/katana. The USDC/moonpay and USDT/oft legs look correctly wired (routers enrolled, addresses checksum-clean, BSC scale math is right: `1/1e12` = 10^(18-6)). A few things worth addressing, flagged inline.

**Most important:** the CROSS ctusd-usdc-ironbridge config now advertises a full spoke-to-spoke mesh, but the deploy wiring is hub-and-spoke through citrea only (every spoke has `destinationConfigs` to `4114` exclusively; only citrea fans out to all spokes). Those spoke-to-spoke connections don't exist on-chain. This looks like the main driver of the `CROSS/ctusd-usdc-ironbridge On-Chain ❌` from check-warp-deploy (the report also cites a polygon export failure, so likely not the sole cause).

The `USDT/moonpay On-Chain ❌` appears to be an expected in-progress state rather than a config bug: routers are deployed and enrolled, but the new ISM/hook/fee/rebalancer config hasn't been applied via governance yet (USDC/moonpay is ✅ because its batch already executed). Worth confirming the pending Safe batch lands so the check clears.

Other notes:
- Token blocks in `USDC/moonpay-config.yaml` and `USDT/moonpay-config.yaml` aren't alphabetized by `chainName` (bsc/katana appended at the end). Minor, flagged inline.
- Changeset only mentions the moonpay extension; consider noting the CROSS ironbridge, USDT/oft, and katana metadata changes too.

Leaving Paul's katana `transactionOverrides` question inline since it's still open.

deployments/warp_routes/CROSS/ctusd-usdc-ironbridge-config.yaml
```diff
@@ -5,7 +5,10 @@ tokens:
     coinGeckoId: usd-coin
     collateralAddressOrDenom: "0xaf88d065e77c8cC2239327C5EDb3A432268e5831"
     connections:
+      - token: ethereum|base|0x4adAf07A84B32F86288e05c4f4A09D830fcf8B71
```
paulbalaji: These spoke-to-spoke connections don't match the on-chain topology. `ctusd-usdc-ironbridge-deploy.yaml` is strictly hub-and-spoke: arbitrum/base/ethereum/polygon each have `destinationConfigs` only to citrea (`4114`), and only citrea fans out to all spokes (`1`/`137`/`42161`/`8453`). For this `collateralDepositAddress` route, connectivity is defined by `destinationConfigs`, which has no spoke-to-spoke entries.

Before this PR each spoke connected to citrea only; the PR widened every spoke to a full mesh (e.g. arbitrum now lists base/ethereum/polygon). Those legs aren't deployed, which is likely why check-warp-deploy reports On-Chain ❌ here.

Suggest reverting spokes to citrea-only and keeping just the new polygon↔citrea leg (which is correctly backed by deploy: polygon `destinationConfigs.4114` + citrea `destinationConfigs.137`). The citrea block adding polygon is correct.

deployments/warp_routes/USDT/oft-config.yaml
```diff
@@ -2,29 +2,46 @@
...
     standard: EvmHypCollateral
     tokenType: collateralOft
     symbol: USDT
   - addressOrDenom: "0x93bfFA2231fbA5029997603fb68D9aC08024Baa2"
     chainName: plasma
+    collateralAddressOrDenom: "0xb8ce59fc3717ada4c02eadf9682a9e934f625ebb"
     connections:
       - token: ethereum|arbitrum|0xE4c1A1E54C232454311cF68610c3885FdD991c0E
       - token: ethereum|ethereum|0x7Bb4fe8f406fB7B22487Fa2e3BCbCb6A38cF29E6
+      - token: ethereum|polygon|0x361A7DA84A1f4A0c565772C45Ed838C6a0fAea94
```
paulbalaji, line 33: The plasma entry is missing `tokenType: collateralOft`. The arbitrum, ethereum, and polygon entries all have it, and `oft-deploy.yaml` sets plasma `type: collateralOft`, so this is an inconsistency in the plasma block (CodeRabbit's earlier thread is marked resolved but the field is still absent at this head). Add `tokenType: collateralOft` to keep parity.

Minor: in the entries that do have it, `tokenType` is placed after `standard` but before `symbol`, so keys aren't alphabetical (`symbol` should precede `tokenType`).

chains/katana/metadata.yaml
```diff
@@ -24,3 +24,6 @@ protocol: ethereum
 rpcUrls:
   - http: https://rpc.katana.network
```
paulbalaji, line 25: Echoing Paul's open question: putting `transactionOverrides` in registry chain metadata applies it to all consumers (SDK, CLI, integrators), not just the agents. If this gas cap is only needed for agent submission, it likely belongs in monorepo agent config (`rust/main/config/`) rather than shared metadata.

Also worth a sanity check: `maxFeePerGas == maxPriorityFeePerGas == 0.01 gwei` leaves zero headroom for base fee. Fine on a near-zero-base-fee L2, but if katana's base fee ever rises, txs become underpriced and can stall. Consider `maxFeePerGas > maxPriorityFeePerGas`.

deployments/warp_routes/USDC/moonpay-config.yaml
```diff
@@ -1,72 +1,141 @@
...
-    name: XO Cash
-    standard: SealevelHypCrossCollateral
-    symbol: XO
+    symbol: vbUSDC
+    name: Vault Bridge USDC
+    addressOrDenom: "0x936e8A1fBD8317Be59A9B8924a300993c8Bf7ce6"
+    collateralAddressOrDenom: "0x203A662b0BD271A6ed5a60EdFbd04bFce608FD36"
+    logoURI: /deployments/warp_routes/USDC/logo.svg
+    connections:
+      - token: ethereum|arbitrum|0xeBC079D41C41a0ef7e54aa7Af867df9a621C9bE0
```
paulbalaji: Minor: token blocks aren't alphabetized by `chainName` (bsc/katana are appended at the end rather than sorted in). Per repo YAML ordering this should be arbitrum, base, bsc, citrea, ethereum, katana, polygon, solanamainnet. Same applies to `USDT/moonpay-config.yaml`.

chains/katana/metadata.yaml
troykessler: needed this specifically for cli, still dont know why but katana would just quote me a gas price of 20 gwei and drain my wallet without this addition

deployments/warp_routes/CROSS/ctusd-usdc-ironbridge-config.yaml
troykessler: removed those connections, those were autogenerated by the cli

paulbalaji (commented):
## Follow-up review

Re-reviewed at `2b32ecd8`. All earlier findings are addressed:

- **CROSS ctusd-usdc-ironbridge**: the spoke-to-spoke mesh is removed - each spoke now connects to citrea only, and citrea fans out to all spokes, matching the hub-and-spoke `destinationConfigs` in the deploy file. This is what flipped the warp-deploy check to Config Sync ✅.
- **USDT/oft-config.yaml**: plasma now has `tokenType: collateralOft`.
- **moonpay configs**: token blocks are alphabetized by `chainName`.

I also checked the new ownership / gas commits and they look correct:
- bsc token owner → 4/9 BSC Safe; katana token owner → valid interchain-account clone of the katana `interchainAccountRouter`.
- ethereum fee ownership moved to a real 4/9 Safe (the previous `0x89d2…` value was not a Safe).
- katana `transactionOverrides` confirmed needed for CLI gas quoting; reasonable to keep in metadata.

The remaining `On-Chain ❌` on the moonpay legs looks like the expected pre-governance state (routers enrolled, ISM/hook/fee config not yet applied on-chain). Worth confirming the governance batch lands so the check clears.

Non-blocking nit: USDT moonpay display `name`s are inconsistent across legs ("USD₮0" / "Tether USD" / "USDT0" / "Vault Bridge USDT"), carried over from the original deploy. Cosmetic only.

LGTM on the config side, thanks!

paulbalaji (approved):
Approving the config. Verified independently:

- All six new-chain collateral token addresses checked on-chain (`cast`): bsc USDC/USDT (18 dec, scale 1/1e12 correct), katana vbUSDC/vbUSDT (6 dec), polygon USDC/USDT0 (6 dec) — symbols and decimals match config.
- Owners resolve correctly: bsc → 4/9 BSC Safe, katana → valid interchain-account clone, ethereum fee → real 4/9 Safe.
- CROSS ironbridge restored to hub-and-spoke (matches deploy `destinationConfigs`); oft plasma `tokenType` and moonpay alphabetization fixed.

Note: `check-warp-deploy` On-Chain ❌ on the moonpay legs is the expected pre-governance state (config sync is ✅) — the ISM/hook/fee governance batch needs to land for it to clear, so hold the actual merge until that executes / the check goes green.

Linear context: AW-644 (CROSS/moonpay polygon+bsc) and AW-437 (USDC FPWR to BSC+katana) confirm intent.

PR #1556, opened 2026-06-08
feat: extend USDT/eni with tron

troykessler:
### Description

Tron extension for USDT/eni, will make pr ready for review once txs have been executed


Mo-Hussain (approved):
Can we check why the check warp deploy job is failing?

troykessler:
> Can we check why the check warp deploy job is failing?

we manually limited the connections in the local config so that nexus only shows bsc, eni and tron connections, on chain however we have the full connection matrix, thats why its failing in this special case here

PR #1563, opened 2026-06-16
feat: limit USDT/eni Nexus connections to bsc + tron

haggis-hyperlane:
Stacked on top of #1556 (`feat/eni-tron-extension`) — for @Troy to consider.

## Context
#1556 extends `USDT/eni` with the Tron leg, but it also rewrites the displayed `connections` graph for every leg into a **full 8-leg mesh** and relabels arbitrum/polygon to `USD₮0`/`USDT0`. Per eda ("we should only add Tron thats new addition here") and Nam's review note, the Nexus-visible change should be the Tron addition only.

## What this does
Trims `eni-config.yaml` `connections` back to the curated subset so that, from the eni synthetic, the visible collateral chains go `[bsc]` → `[bsc, tron]` — matching what `main` already does (it intentionally under-declares connections relative to the on-chain mesh).

**Net effect vs `main` is exactly "main + Tron":**
- eni gains a single `tron|tron|0x34662Cb7…` connection (keeps existing bsc).
- New `tron` leg added (`TronHypCollateral`), connected to eni only.
- The other 5 EVM legs (arbitrum/base/ethereum/optimism/polygon) stay non-routable in Nexus, as today.
- Reverts the cosmetic `USD₮0`/`USDT0` relabels.

## What this does NOT touch
- `eni-deploy.yaml` (mailboxes, `RoutingFee`/`OffchainQuotedLinearFee` 5bps blocks) — left as #1556 has it; not consumed by the Nexus UI.
- On-chain enrollments — the route remains a true full mesh on-chain (verified via fork sim). This is a **display-only** curation.

## Decision for the reviewer
- If the goal is *Nexus shows bsc + tron from eni* → merge this on top of #1556.
- If the goal is *Nexus surfaces the real on-chain full mesh* → merge #1556 as-is and close this.

Linear: [AW-675](https://linear.app/hyperlane-xyz/issue/AW-675/eni-usdt-tron-expansion)

### DIFF
--- a/deployments/warp_routes/USDT/eni-config.yaml
+++ b/deployments/warp_routes/USDT/eni-config.yaml
@@ -4,17 +4,9 @@ tokens:
     chainName: arbitrum
     coinGeckoId: tether
     collateralAddressOrDenom: "0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9"
-    connections:
-      - token: ethereum|base|0x3a669c00F7A041330646AC8AE89d74e104bd1363
-      - token: ethereum|bsc|0xC05617bc2490CB57a8db163b5551bb7E532695e3
-      - token: ethereum|eni|0x545E289B88c6d97b74eC0B96e308cae46Bf5f832
-      - token: ethereum|ethereum|0x291A485Fd288786728Ba34052b95DDb60278AB49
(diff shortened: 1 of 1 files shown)


troy:
@haggis-hyperlane: you’re @ mentioning the wrong Troy. I’m not associated with this project.

troykessler (approved):

PR #1604, opened 2026-07-16
feat: usdc seismictestnet warp route

troykessler:
## Summary
- Add USDC warp route between sepolia (collateral) and seismictestnet (synthetic)
- Add predicate wrapper config on the sepolia leg
- Add coinGeckoId and logoURI metadata for the token
- Add seismictestnet transactionOverrides (gasLimit, deploymentGasLimitFallback)

## Test plan
- [ ] `pnpm run lint`
- [ ] `pnpm run test:unit`


chains/seismictestnet/metadata.yaml
```diff
@@ -24,3 +24,6 @@ protocol: ethereum
 rpcUrls:
   - http: https://testnet-1.seismictest.net/rpc
 technicalStack: seismic
+transactionOverrides:
+  gasLimit: 500000
+  deploymentGasLimitFallback: 5000000
```
paulbalaji: these would get propagated to the agent configs too - is that intentional?

troykessler: removed them again since they are not necessary anymore

paulbalaji (approved):
Reviewed at 969d9a3. Prior concern addressed; build, lint, test, format, and changeset checks pass. The check-warp-deploy failure is hook derivation while route config sync passes.

PR #1610, opened 2026-07-18
fix: deprioritize rate-limited base.drpc.org in Base rpcUrls

haggis-hyperlane:
## Summary
- Base was the only remaining chain in the NES warp route with `base.drpc.org` as its primary RPC after #1609 fixed BSC.
- Nexus's viem `fallback()` transport uses `rank=false`, so it re-selects the rate-limited primary on every request instead of demoting it — causing recurring HTTP 429 errors on the base leg of NES transfers (NESA still reporting the same error post-BSC-fix).
- Reordered `chains/base/metadata.yaml` to promote healthy public endpoints (`mainnet.base.org`, `base-pokt.nodies.app`, `base.public.blockpi.network`) above `base.drpc.org`; kept the currently-down `base.llamarpc.com` last.

All three promoted endpoints verified 200 and in sync (block 0x2e847d8) at time of writing.

## Test plan
- [ ] CI green (changeset + schema validation)
- [ ] NESA confirms Nexus NES transfers succeed once merged and propagated (~5min GitHub CDN cache)

### DIFF
--- a/.changeset/base-rpc-reorder-drpc.md
+++ b/.changeset/base-rpc-reorder-drpc.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Reordered Base rpcUrls to prioritize healthy public endpoints (mainnet.base.org, base-pokt.nodies.app, base.public.blockpi.network) over the rate-limited base.drpc.org, mirroring the earlier BSC reorder.
--- a/chains/base/metadata.yaml
+++ b/chains/base/metadata.yaml
@@ -32,10 +32,10 @@ nativeToken:
   symbol: ETH
 protocol: ethereum
 rpcUrls:
-  - http: https://base.drpc.org
   - http: https://mainnet.base.org
-  - http: https://base.public.blockpi.network/v1/rpc/public
-  - http: https://base.llamarpc.com
-  - http: https://1rpc.io/base
   - http: https://base-pokt.nodies.app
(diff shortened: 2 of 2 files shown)


Skunkchain (changes requested):
Blocking finding in `chains/base/metadata.yaml:35`: this makes `https://mainnet.base.org` the primary RPC, but Base explicitly documents that endpoint as rate limited and not suitable for production systems: https://docs.base.org/base-chain/quickstart/connecting-to-base

The PR explains that Nexus uses `rank=false` and therefore re-selects the first URL on every request. Promoting another explicitly rate-limited endpoint can reproduce the same failure mode this change is intended to resolve. Please put one of the healthy provider endpoints first, preferably `https://base-pokt.nodies.app`, followed by BlockPI, and retain `mainnet.base.org` as a later fallback.

Live validation from a browser-origin request confirms Nodies and BlockPI both return Base chain ID `0x2105`, current block height, HTTP 200, and valid CORS. The dRPC demotion and dead LlamaRPC placement are otherwise correct, and all CI checks are green.

Skunkchain (approved):
The blocking ordering issue is resolved in 404b2e03: Nodies and BlockPI are now primary and secondary, with Base's documented rate-limited endpoint retained only as a later fallback. Live checks confirmed both primary endpoints return Base chain ID 0x2105, current blocks, HTTP 200, and browser-compatible CORS. All CI checks are green. No remaining blocking findings.

PR #1627, opened 2026-07-27
fix(warp): add solanamainnet leg to DIME/paradex deploy config

haggis-hyperlane:
## Summary

Adds the missing `solanamainnet` leg to the DIME/paradex warp route deploy config. The route deploys to Solana but the deploy config lacked the solana block, so `check-warp-deploy` had no expected config to compare against (and no gas config for the solana leg).

## Verification

All fields for the solana synthetic were read directly from on-chain state via `SvmSyntheticTokenReader`:

- `owner` = `HBPwc1dSuaJCEwWkJvfeWUqJguFqPTVaggfDGssc3LVt` — verified on-chain (exact match, not inferred)
- `decimals` = 9 — verified on-chain
- `interchainSecurityModule` = null on-chain, so this leg does NOT hit the SVM multisig-ISM reader stub
- `type` = synthetic

`token` (mint PDA) is omitted to match the sibling synthetic blocks (hyperevm, mode) and because `token` is excluded from the warp-check comparison for synthetic token types. `name`/`symbol` are not part of the comparison either.


nambrot (approved):

deployments/warp_routes/DIME/paradex-deploy.yaml
```diff
@@ -25,3 +25,10 @@ paradex:
   symbol: DIME
   token: "0x316cb56a8cc7c478ea7c9b22f33571330bc61ea0312f2222d182b4b6b3fa4b2"
   type: collateralDex
+solanamainnet:
+  decimals: 9
```
paulbalaji, line 40: **High:** This new Solana leg uses 9 decimals while the rest of the route uses 18, so it needs `scale: 1000000000`. Without it, the deploy config cannot represent the route's 18→9 conversion; the exact-head `check-warp-deploy` run reports `Found invalid or missing scale for inconsistent decimals`. Existing 18/9 Solana routes use the same `10^9` scale.

paulbalaji (commented):
Reviewed at `1c267eec12b07b63bb9e78a51d499d3df02ee4dd`.

The new owner is corroborated by the monorepo's Paradex Solana configuration, and build/lint/format/unit/changeset checks pass. The exact-head warp-deploy check still rejects the added leg because its decimal conversion is incomplete; finding inline. The existing provenance-documentation thread also remains unresolved. No `REQUEST_CHANGES` review submitted.

deployments/warp_routes/DIME/paradex-deploy.yaml
haggis-hyperlane: Addressed in f57db205 / 28520cb3: added `scale: 1000000000` to the solanamainnet leg in both the deploy config and the warp core config (10^(18-9), matching the existing 18/9 Solana routes e.g. CDX, HOLO). The scale-consistency test and the exact-head check both needed the deploy+core scales to match.

deployments/warp_routes/DIME/paradex-deploy.yaml
```diff
@@ -25,3 +25,10 @@ paradex:
...
+  gas: 300000
+  name: DIME
+  owner: HBPwc1dSuaJCEwWkJvfeWUqJguFqPTVaggfDGssc3LVt
+  symbol: DIME
+  type: synthetic
```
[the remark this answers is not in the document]
haggis-hyperlane: Provenance: `HBPwc1dSuaJCEwWkJvfeWUqJguFqPTVaggfDGssc3LVt` is the established Paradex Solana operator owner — the same address already used for the solanamainnet leg of the merged `USDC/paradex` route (`deployments/warp_routes/USDC/paradex-deploy.yaml`). It's not a new key; it's the shared Paradex Solana owner, which is why @paulbalaji corroborated it against the monorepo's Paradex Solana config. Registry deploy configs don't carry inline provenance comments by convention, so leaving the value as-is.

paulbalaji (approved):
Reviewed exact head `1058e4252e87b68d243153a2182a6a24e878a066`, including the author replies. The 18→9 conversion is now consistently represented as `scale: 1000000000` in deploy and core config. I independently read the live routers: Solana owner/scale/full remote set match; HyperEVM's intentionally partial enrollment and 68k/68k/5m gas match; Ethereum and Mode both route to Solana with 300k gas. The established Paradex Solana owner also matches the merged USDC route.

The red `check-warp-deploy` is not a config-sync regression: config sync passes, while the on-chain phase surfaces the existing Paradex reader/state discrepancy (`collateralDex` read as `collateral`, destination gas read as zero) plus the synthetic mint readback. That follow-up should remain visible, but it does not invalidate these additions. Build/lint/format/test/changeset checks pass and YAML ordering is correct.

PR #1629, opened 2026-07-28
chore: rotate dead public RPC endpoints

troykessler:
## Summary
- Removed dead/unreachable public RPC entries for `arbitrum`, `base`, `blast`, `bsc`, `polygon`, `viction` and added verified-alive replacements (mostly diversified across PublicNode, Tenderly, and official chain-foundation endpoints, avoiding same-provider duplicates).
- Replaced the sole public RPC for `adichain` and `robinhood` (each had exactly one entry, and it was dead) with a working alternative found via the chain's official docs / `ethereum-lists/chains`.
- Every remaining and newly-added URL was live-tested (`eth_blockNumber` via `ethers.StaticJsonRpcProvider`, with explicit chainId to skip network auto-detection) immediately before this PR, including a re-check right before opening it.
- `oortmainnet`'s sole public RPC (`mainnet-rpc.oortech.com`) is also dead, but no alternative could be found anywhere (official docs, ChainList, rpc.info all list only this one URL) — left unchanged.
- `celo` was investigated (one RPC intermittently flagged as down) but all 3 entries came back alive on careful re-check — no change made.

## Why
Auditing AW's Ethereum-protocol validators' RPC health surfaced several registry-listed "public" RPCs that are now dead (DNS no longer resolves, or the endpoint now requires a paid plan / API key). Dead registry entries are silently included wherever `rpcUrls` is consumed, including a not-yet-shipped validator quorum-verification pool size calculation (hyperlane-monorepo#9124), where dead entries occupying slots in a fixed-size quorum pool can push the pool below its required 2/3 threshold. `bsc` was the one chain where 3 of 5 public RPCs being dead pushed it below threshold; fixed here so the raw registry list stays a safe default.

## Verification
Every URL in the diff (both removed-from and added-to state) was tested for liveness right before submitting. Removed entries failed with a mix of DNS resolution failures (verified against `1.1.1.1` directly, not just a local resolver, to rule out local network artifacts) and HTTP-level rejections (401/403 paywalled, 5xx errors), reproduced consistently across multiple attempts spaced apart in time.


chains/robinhood/metadata.yaml
```diff
@@ -22,5 +22,5 @@ nativeToken:
   symbol: ETH
 protocol: ethereum
 rpcUrls:
-  - http: https://rpc.chain.robinhood.com
-technicalStack: other
+  - http: https://rpc.mainnet.chain.robinhood.com
+technicalStack: arbitrumnitro
```
paulbalaji: Medium: this classification is correct, but it makes `index.from` mandatory. The current `test` job fails while constructing `ChainMetadataManager` with `An index.from value is required for Arbitrum Nitro chains`. Add the Robinhood Mailbox deployment/start block under `index.from`; `combine` passing alone does not produce schema-valid metadata.

chains/bsc/metadata.yaml
```diff
@@ -26,8 +26,7 @@ nativeToken:
 protocol: ethereum
 rpcUrls:
   - http: https://bsc.blockrazor.xyz
-  - http: https://bsc-pokt.nodies.app
   - http: https://bsc.drpc.org
-  - http: https://bnb.rpc.subquery.network/public
-  - http: https://binance.llamarpc.com
+  - http: https://bsc-dataseed1.bnbchain.org
```
paulbalaji: Medium: this official BSC endpoint is not a usable generic registry RPC. It rejects even a one-block, Mailbox-address-filtered `eth_getLogs` call with `-32005 limit exceeded`, and BNB documents `eth_getLogs` as disabled on its public endpoints. `test/health/rpc-health.test.ts` explicitly requires a 100-block Mailbox log query for EVM RPCs, so `test:rpc-health-mainnet` will fail. Replace it with an endpoint that supports logs.

paulbalaji (commented):
Reviewed at `fbe58246e05e89dc66f306f993f2fb2d439f6ea4`. The RPC rotations otherwise returned the expected chain IDs and current block heights. The new head fixed the earlier missing changeset and generated Nitro classification, so `changeset` and `combine` now pass. Two actionable findings remain inline; the Robinhood metadata issue is the cause of the current `test` failure.

chains/robinhood/metadata.yaml
troykessler: will revert that

chains/bsc/metadata.yaml
troykessler: Confirmed and fixed. Verified with the actual health-check semantics (100-block, Mailbox-address-filtered eth_getLogs, matching test/health/rpc-health.test.ts) rather than just a basic eth_blockNumber call. Turns out bnbchain.org/defibit.io/ninicoin.io dataseed nodes all reject it the same way, and the pre-existing blockrazor.xyz entry also fails (25-block cap). Replaced with drpc.org, publicnode.com, and blxrbdn.com (bloXroute) -- all three verified to return real logs over a 100-block Mailbox-filtered range.

chains/robinhood/metadata.yaml
```diff
@@ -22,5 +22,5 @@ nativeToken:
   symbol: ETH
 protocol: ethereum
 rpcUrls:
-  - http: https://rpc.chain.robinhood.com
+  - http: https://rpc.mainnet.chain.robinhood.com
 technicalStack: other
```
paulbalaji: Medium: reverting this to `other` cannot satisfy CI—`update-nitro-chains.sh` detects Robinhood as Nitro and rewrites it, so `combine` remains red. The actual Mailbox deployment transaction [`0xe08c…3b08a`](https://robinhoodchain.blockscout.com/tx/0xe08c2a374dd15ebf8b35d7079f5493f5abc41e4c959529709fdd1d8f8fc3b08a) successfully created `0x3a867fCfFeC2B790970eeBDC9023E75B0a172aa7` at block **47247** (`0xb88f`). Please keep `technicalStack: arbitrumnitro` and add:

```yaml
index:
  from: 47247
```

That makes the generated Nitro classification schema-valid, satisfying both `combine` and `test`.

paulbalaji (commented):
Targeted follow-up at `273a98b1c2ed4f15bbc392c9f2a3a6fa6d0bc997`: the BSC finding is fixed. Robinhood still needs the Nitro classification plus its now-confirmed deployment block; reverting the classification is why `combine` remains red.

paulbalaji (approved):
Approved at `adff7ab5f9f00307e55f721193a50d0f99a75ea4`. Both prior findings are resolved: the BSC replacements support the registry's 100-block Mailbox log-health query, and Robinhood is correctly classified as Arbitrum Nitro with the verified Mailbox deployment index `47247`. The Nitro generator leaves the worktree clean, all review threads are resolved, the changeset is present, and CI is fully green.

PR #1636, opened 2026-08-04
chore: upgrade Node 26 and TypeScript 7 toolchain

paulbalaji:
## Summary

- upgrade Node to 26.6.0, pnpm to 11.20.0, TypeScript to 7.0.2, oxlint to 1.77.0, and oxfmt to 0.62.0
- upgrade checkout, setup-node, cache, and pnpm setup actions to current majors
- refresh compatible runtime and development dependencies, including Hyperlane SDK/utils 40
- move Pino 8 from devDependencies to dependencies so exported Logger types resolve consistently for consumers
- regenerate the warp schema with SDK 40 Sealevel ALT options
- keep Zod on v3 and Pino on v8 to preserve the published API

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm build`
- `pnpm lint`
- oxfmt 0.62 format check
- unit tests — 8,169 passing, 1 pending

Includes a patch changeset for the published dependency and schema changes.


package.json
```diff
@@ -47,35 +47,35 @@
...
-    "zod-to-json-schema": "^3.22.5"
+    "sinon": "^22.1.0",
+    "svgo": "4.0.2",
+    "tsx": "^4.23.1",
+    "typescript": "7.0.2",
+    "zod-to-json-schema": "^3.25.2"
   },
   "engines": {
-    "node": ">=24"
+    "node": ">=26"
```
troykessler: This raises the published package's supported runtime from Node 24 to Node 26, but the changeset is only `patch`. Consumers running with `engine-strict` (or package managers/environments that enforce `engines`) can no longer install/use a patch release on Node 24, so this is a breaking package contract change. Either keep `engines.node` at `>=24` if the library still works there, or make the changeset major.

troykessler (commented):
Found one release-compat issue.

package.json
paulbalaji: This finding was correct and is fixed in [f28bd1cc](https://github.com/hyperlane-xyz/hyperlane-registry/commit/f28bd1cc6e1f9fba6c1fc5d30d4e3bed1b2cdec1): the published package keeps engines.node at >=24, while development and CI remain on Node 26. I verified the upgraded toolchain on Node 24.13.0 with a successful build and 8,169 passing unit tests (1 pending); the exact pushed head has 12 completed checks with zero failures.

troykessler (commented):
Reviewed at `f28bd1cc6e1f9fba6c1fc5d30d4e3bed1b2cdec1`.

Follow-up review. No new actionable findings. The prior engine-compat issue was fixed by keeping the published package on `engines.node >=24` while using Node 26 for development/CI.

troykessler (commented):
Reviewed at `f28bd1cc6e1f9fba6c1fc5d30d4e3bed1b2cdec1`.

No blocking findings from the diff. Focused on published dependency surface, generated schema changes, action/runtime bumps, and the Pino runtime dependency move.

troykessler (approved):

PR #1646, opened 2026-08-10
chore: kiichain proxy admin renouncement

troykessler:
### Description

Kiichain proxy admin renouncement of synthetic legs


troykessler:
instead of renouncing we are deploying timelocks: https://github.com/hyperlane-xyz/hyperlane-registry/pull/1649

PR #1661, opened 2026-08-20
fix: add collateral metadata to CCTP v2 routes

Xaroz:
### Description

- Backfill `collateralAddressOrDenom` from each chain's CCTP V2 deploy config for the fast and standard USDC routes.
- Add `coinGeckoId: usd-coin` to every route leg.
- Regenerate the combined warp route config.

### Backward compatibility

Yes. This adds missing token metadata without changing deployed route addresses or behavior.

### DIFF
--- a/.changeset/backfill-cctp-v2-collateral-metadata.md
+++ b/.changeset/backfill-cctp-v2-collateral-metadata.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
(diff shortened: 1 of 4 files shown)


paulbalaji (approved):
Validated all 28 CCTP V2 fast/standard routers: live wrappedToken() values match the added collateral addresses. Generated output reproduces cleanly; 8,228 unit tests pass. The failed deploy check is limited to a Sei explorer API-key error, while config sync passes.

PR #1664, opened 2026-08-24
feat: secure METIS LazAI route

troykessler:
## Summary

- require fallback, pausable, and rate-limited ISMs on both route legs
- cap inbound transfers at approximately 100,000 METIS per day
- configure one-day timelocks as ProxyAdmin owners

## Validation

- deploy config schema validation
- transaction simulations on Metis and LazAI
- git diff whitespace validation

### DIFF
--- a/.changeset/secure-metis-lazai-route.md
+++ b/.changeset/secure-metis-lazai-route.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Add fallback, pausable, and daily 100,000 METIS rate-limited ISMs to both METIS/lazai legs, and configure one-day timelocks to own their proxy admins.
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
Reviewed exact head b6262a24. No blocking issues found. The 3-of-3 fallback + pausable + rate-limited ISM composition preserves mailbox-default authentication while enforcing pause and ~100,000 METIS/day token-bucket controls on both inbound legs; the one-day timelock configuration correctly targets ProxyAdmin ownership. Exact-head build, warp-route tests (1,834 passing), changeset status, diff check, and linked config export passed. Live reads confirmed both routers/ProxyAdmins remain owned by the configured EOA and both mailbox default ISMs are deployed. The failing on-chain check is expected pre-deployment drift: it probes getMinDelay() on the current EOA before the target timelocks exist.

PR #1673, opened 2026-08-26
feat: cut over NES warp route to Nesachain

haggis-hyperlane:
## Summary

- cut the NES route over from `nesa` (`41443`) to `nesachain` (`41444`)
- add the live Nesachain native router `0xbb0AE51BCa526cF313b6a95BfaB020794af6C394`
- update the synthetic legs to the aggregation 2/2 ISM
- remove the old paused routing hooks

## Status

The Nesachain core deployment is on `main` via #1679, and the native leg is live and owned by the Ethereum ICA. Hyperlane enrollment multisig transactions are proposed and in progress. The existing EVM and Solana owner transactions can execute in parallel; transfers should wait until both sides are live.

`check-warp-deploy` is expected to reconcile after the existing-leg owner transactions execute.

## Validation

- registry build, lint, formatting, changeset, and focused `NES/bsc` tests
- live Nesachain router, mailbox, ISM, ownership, and zero-supply backing readback
- decoded EVM batch and replayed Solana transactions against the target config

### DIFF
--- a/deployments/warp_routes/NES/bsc-config.yaml
+++ b/deployments/warp_routes/NES/bsc-config.yaml
@@ -1,104 +1,111 @@
 # yaml-language-server: $schema=../schema.json
 tokens:
-  - addressOrDenom: "0xEf9295afCff293956E8B149b33449f246f6F107D"
-    logoURI: /deployments/warp_routes/NES/logo.svg
+  - addressOrDenom: "0x8cAd676141FeadB6aCA5eF661E835d5957644033"
     chainName: arbitrum
     connections:
-      - token: ethereum|base|0x87ea09Fe8d9dC6115086F5E0A30CA6750a997f1C
-      - token: ethereum|bsc|0x3131f6B80C26936aB03F7d9D29Eb4Ddf36AC3FB5
-      - token: ethereum|ethereum|0x230f1E241C621d5af670Dad83ebCdd18971E2995
-      - token: ethereum|hyperevm|0xD1B6443d49156B66E7bd8a37673511BE68BFa459
-      - token: ethereum|nesa|0xEEC6574eAbBa52bac3f0277F2cD5Ac7e67197886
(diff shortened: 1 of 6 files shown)


deployments/warp_routes/NES/bsc-config.yaml
```diff
@@ -1,75 +1,5 @@
 # yaml-language-server: $schema=../schema.json
 tokens:
-  - addressOrDenom: "0xEf9295afCff293956E8B149b33449f246f6F107D"
```
paulbalaji, line 3: **[High] Restore a self-consistent core config before merge.** Deleting the synthetic entries leaves the retained `nesa` token connected to tokens no longer present, and removing the Solana entry removes the scale that still exists in `bsc-deploy.yaml`. On this exact head, `WarpCore.FromConfig` throws (`Connected token not found: arbitrum 0xEf9295afCff293956E8B149b33449f246f6F107D`) and the targeted suite has 3 failures, including the Solana scale mismatch. Please regenerate this file with the new addresses/connections/scale after redeploy, or retain the current entries until then; merging this interim file would publish an unusable `NES/bsc` route.

paulbalaji (commented):
Reviewed exact head 9f933489f3345187b01e51404f848a22bb936381. One High finding inline: removing the synthetic core entries leaves the retained Nesa connections dangling and removes the Solana scale, reproducing three NES/bsc unit failures. The ISM topology itself is sound: each 2-of-2 requires both the pausable branch and Mailbox fallback verification; build/schema validation, diff check, and changeset status passed. Posting COMMENT because the current core config is an invalid pre-deployment intermediate state.

paulbalaji (approved):
Reviewed exact head e5447544760e580fb42a9a4aefa60fdc0f628bd6. Prior core-config finding is resolved. Registry CI and check-warp-deploy pass; live standard and ICA-aware warp checks report no violations; the Nesachain RPC returns domain 41444; NES/bsc remains absent from the active Nexus/URE allowlist. Approved for merge.

PR #1674, opened 2026-08-27
feat(warp): pause USDC/aleo transfers to aleo on all EVM legs

haggis-hyperlane:
## Summary
- Added a `fallbackRoutingHook` on each EVM leg of the USDC/aleo route (arbitrum, avalanche, base, bsc, ethereum, optimism, polygon).
- The hook routes the `aleo` destination to a `pausableHook` with `paused: true`, so transfers to aleo revert on dispatch immediately.
- All other destinations fall through to `defaultHook` (mailbox default), so no other legs change behavior.

## Notes
- Config-only change to `aleo-deploy.yaml`. On-chain effect requires a separate `hyperlane warp apply` run to deploy the hooks and wire them onto the token contracts.
- Hook owner per leg matches the existing warp token owner on that chain.
- The solanamainnet (SVM) leg is not included: Sealevel warp routes do not support this EVM domain-routing/pausable hook config, so Solana→aleo transfers are not blocked by this change.

## Test plan
- [ ] Registry CI (schema + sort lint) passes
- [ ] `hyperlane warp apply` dry-run to confirm the 7 pausable hooks deploy and route aleo correctly

### DIFF
--- a/deployments/warp_routes/USDC/aleo-deploy.yaml
+++ b/deployments/warp_routes/USDC/aleo-deploy.yaml
@@ -264,6 +264,21 @@ solanamainnet:
   mailbox: E588QtVUvresuXq2KoNEwAmoifCzYGpRBdHByN9KQMbi
   name: USD Coin
   owner: ABJnd4eWexNte9GYy21ud5hvSwFKWedveP6GCFxXKkCw
+  remoteRouters:
+    "1":
+      address: "0x78Ac7FECD1857f5BEEe98AB39096c9781F976D97"
+    "10":
+      address: "0x1FdA66FA15A261F01F1E09228D41bD0A806d7529"
+    "56":
+      address: "0x5284D803a4563DC5eE83feA80c688b096d70eb75"
+    "137":
+      address: "0x1FdA66FA15A261F01F1E09228D41bD0A806d7529"
+    "8453":
+      address: "0xB46930ca998587A95D9Ee000FA73A071ADD56B64"
+    "42161":
+      address: "0x1FdA66FA15A261F01F1E09228D41bD0A806d7529"
+    "43114":
+      address: "0x1FdA66FA15A261F01F1E09228D41bD0A806d7529"
   scale: 1000000000000
   symbol: USDC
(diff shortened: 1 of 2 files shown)


paulbalaji (commented):
One High functional finding inline. The config shape, Aleo domain mapping, and all seven owners otherwise check out. I also confirmed the red check-warp-deploy result is expected pre-apply drift (all seven live router hooks are currently zero); Config Sync and normal CI pass.

deployments/warp_routes/USDC/aleo-deploy.yaml
```diff
@@ -27,6 +27,16 @@ arbitrum:
       - bridge: "0x4c19c653a8419A475d9B6735511cB81C15b8d9b2"
       - bridge: "0xE086378F7f0afd5C3ff95E10B5e7806a0901b33f"
   decimals: 6
+  hook:
+    domains:
+      aleo:
+        owner: "0x63C65aFC66C7247a3d43197744Da7F5838ACbf77"
+        paused: true
```
paulbalaji, line 34: **High — `paused: true` is not applied by the released deployment path.** The registry pins `@hyperlane-xyz/sdk` 41.3.0, whose `EvmHookModule.deployPausableHook` deploys the hook and transfers ownership without calling `pause()`. All seven live routers currently have `hook() == address(0)`, so `warp apply` takes this fresh-deploy path and installs an unpaused Aleo child; transfers remain enabled. [hyperlane-monorepo#9351](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/9351) is the open, unreleased fix. Please gate rollout on a CLI/SDK release containing that fix (or coordinate explicit owner-governed `pause()` transactions), then verify `paused()` on all seven child hooks.

paulbalaji (approved):
Registry config is correct. The PausableHook initialization issue is in deployment tooling and is tracked by hyperlane-monorepo#9351; treat the fixed CLI/explicit pause action as a rollout prerequisite, not a defect in this PR. Verified Aleo domain mapping, fallback behavior, all seven live owners, Config Sync, and normal CI.

PR #1676, opened 2026-08-28
feat: migrate Starknet and Paradex RPC URLs to JSON-RPC v0_9

haggis-hyperlane:
## Summary

Alchemy is removing the Starknet Mainnet JSON-RPC **v0_8** endpoint on **September 1**. This moves the Starknet and Paradex RPC URLs from the v0_8 spec path to **v0_9**:

- `starknet` (mainnet): lava `→ /rpc/v0_9`
- `paradex` (mainnet): all 3 hosts (`rpc.api`, `juno`, `pathfinder`) `→ /rpc/v0_9`
- `paradexsepolia`: pathfinder `→ /rpc/v0_9`
- `chains/metadata.yaml` regenerated from the per-chain metadata (auto-gen aggregate)

Verified live: starknet lava `/rpc/v0_9` and paradex `rpc.api.prod` `/rpc/v0_9` both return `starknet_specVersion = 0.9.0`.

## Coordination note
These chains run on `protocol: starknet`, whose clients pin one JSON-RPC spec. Consumers must be able to speak v0.9 before/with this flip:
- **Rust agents** (relayer/validator/scraper): upgraded to `starknet-rs 0.17` (v0.9) in hyperlane-monorepo#9355.
- **TypeScript stack** (`starknet.js 7.6.4` = RPC 0.8 only — SDK/CLI/widgets/Nexus): does **not** yet speak v0.9. A `starknet.js` v8 bump is required for those consumers before they can use these URLs without breaking. Merge/rollout should be sequenced accordingly.

## Not included
- `starknetsepolia`: no reliable public `/rpc/v0_9` endpoint currently (lava testnet down, blast dead, publicnode only serves a version-agnostic bare path at 0.10.2). Left as-is.

### DIFF
--- a/chains/paradex/metadata.yaml
+++ b/chains/paradex/metadata.yaml
@@ -24,7 +24,7 @@ nativeToken:
   symbol: FUEL
 protocol: starknet
 rpcUrls:
-  - http: https://rpc.api.prod.paradex.trade/rpc/v0_8
-  - http: https://juno.api.prod.paradex.trade/rpc/v0_8
-  - http: https://pathfinder.api.prod.paradex.trade/rpc/v0_8
+  - http: https://rpc.api.prod.paradex.trade/rpc/v0_9
+  - http: https://juno.api.prod.paradex.trade/rpc/v0_9
+  - http: https://pathfinder.api.prod.paradex.trade/rpc/v0_9
 technicalStack: other
--- a/chains/paradex/metadata.yaml
+++ b/chains/paradex/metadata.yaml
@@ -25,6 +25,4 @@ nativeToken:
 protocol: starknet
 rpcUrls:
   - http: https://rpc.api.prod.paradex.trade/rpc/v0_9
-  - http: https://juno.api.prod.paradex.trade/rpc/v0_9
-  - http: https://pathfinder.api.prod.paradex.trade/rpc/v0_9
 technicalStack: other
--- a/.changeset/starknet-paradex-rpc-v0_9.md
+++ b/.changeset/starknet-paradex-rpc-v0_9.md
@@ -2,4 +2,4 @@
 '@hyperlane-xyz/registry': patch
 ---
 
-The Starknet and Paradex RPC URLs are migrated from the JSON-RPC v0_8 spec path to v0_9 (starknet mainnet lava, paradex mainnet, and paradexsepolia pathfinder) ahead of the Alchemy Starknet v0_8 endpoint removal on September 1.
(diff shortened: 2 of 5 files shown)


chains/paradex/metadata.yaml
```diff
@@ -24,7 +24,5 @@ nativeToken:
   symbol: FUEL
 protocol: starknet
 rpcUrls:
-  - http: https://rpc.api.prod.paradex.trade/rpc/v0_8
-  - http: https://juno.api.prod.paradex.trade/rpc/v0_8
-  - http: https://pathfinder.api.prod.paradex.trade/rpc/v0_8
+  - http: https://rpc.api.prod.paradex.trade/rpc/v0_9
```
paulbalaji, line 27: P1: This cannot be the only Paradex RPC yet. Live `starknet_getTransactionReceipt` and `starknet_getBlockWithReceipts` calls against `/rpc/v0_9` consistently return `-32000 json unmarshal failed: ... l1_data_gas ... string ... into ... int`; identical requests succeed on `/rpc/v0_8`. The Starknet agent provider uses `get_transaction_receipt`, so this would break transaction processing after the v0.9 client rollout. Please retain a working receipt-capable endpoint, or have Paradex fix and reverify this endpoint before cutover.

paulbalaji (commented):
Reviewed exact head `29eb44e24435c9388bbdd54c5f56701549a30d73`. The retained URLs report the expected chain IDs and RPC 0.9.0; generated metadata, changeset, formatting, tests, and exact-head CI are clean. I cannot approve the cutover yet for two operational reasons.

1. Paradex `/rpc/v0_9` currently fails receipt methods used by the Rust agent provider; see inline.
2. The existing [TypeScript compatibility finding](https://github.com/hyperlane-xyz/hyperlane-registry/pull/1676#discussion_r3879051437) remains valid despite its thread being resolved without a reply or implementation. This registry resolves starknet.js 7.6.2 and the monorepo resolves 7.6.4, while the [official compatibility table](https://docs.starknet.io/learn/cheatsheets/compatibility) maps RPC 0.9 to starknet.js v8. PR #9355 itself says registry publication is gated on that separate consumer migration.

Please retain v0.8-compatible URLs until the Paradex receipt failure is resolved and the TypeScript v8 migration is landed/deployed. The PR body is also stale: the final diff no longer changes Paradex Sepolia or all three Paradex hosts.

chains/paradex/metadata.yaml
troykessler: reached out to the paradex team, i think they have a bug in their rpc endpoint, had already a similar issue with their v0_8 endpoints, lets wait for their response

troykessler: accepted for now, waiting on the fix from paradex

troykessler (approved):

PR #1666, opened 2026-08-24
chore(KII/kiichain): add rate-limit + pausable ISMs on EVM legs

haggis-hyperlane:
## Summary

Sets the target `interchainSecurityModule` on the KII/kiichain warp route's **5 EVM legs** (base, bsc, ethereum, mantle, polygon) to a `staticAggregationIsm` (threshold = all 3 modules) composed of:

- `rateLimitedIsm` — `maxCapacity` `9999999999999999936000000` (10M KII/day), owned by each leg's existing token owner
- `pausableIsm` — `paused: false`, owned by the dedicated Turnkey pauser key `0x60Cc386C85717CB51C2827A75e14826883dF5da4`
- `defaultFallbackRoutingIsm` — preserves the mailbox-default security, owned by each leg's existing token owner

The **kiichain leg is intentionally left unchanged** for now — that chain is currently halted (block height stuck), so no dispatches originate from it and the pauser key isn't funded there yet.

This PR carries the declarative target config; the on-chain wiring lands via `warp apply` (deployer-run). The `check-warp-deploy` CI will go red until apply completes, then green.

| Field | Value |
| ----- | ----- |
| **Linear** | https://linear.app/hyperlane-xyz/issue/AW-752 |
| **Scope** | 5 of 6 legs (kiichain deferred) |

### Apply-time notes for the deployer

- The Sep-incident already deployed **paused** `PausableIsm`s on these 5 legs. Decide whether to reuse them vs. deploying fresh (`paused: false`) ones — a fresh unpaused pausable is what this config declares; reusing the paused ones would deliver a paused route.
- `staticAggregationIsm` modules are immutable, so a new aggregation contract will deploy regardless (adding the rate-limit module).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1681, opened 2026-08-29
chore(KII/kiichain): record kiichain leg ISM in registry

haggis-hyperlane:
## Summary
- PagerDuty Q3F3NBZJLZ8SXV: Warp Config Checker fires `ConfigMismatch` for `interchainSecurityModule` on KII/kiichain (`chain=kiichain`).
- Root cause: #1666 added the staticAggregationIsm to the EVM legs on-chain and in the registry but intentionally omitted the kiichain leg while that chain was halted. The kiichain router has the same ISM on-chain (`0x5dA16E7E42232496c9893cFD419bCd4a71Eb4723`), and kiichain is now live, so the checker flags the missing registry entry.
- Fix: record the on-chain ISM on the kiichain leg. Aggregation matches EVM legs: rateLimited (`maxCapacity=999999999999999993600000`, owner `0x30C48d3F...`), pausable (owner `0x60Cc386C...`, dedicated Turnkey pauser), defaultFallbackRouting (owner `0x30C48d3F...`), threshold 3.

Verified on-chain via `cast` against kiichain RPC: router `0xEEC6574eAbBa52bac3f0277F2cD5Ac7e67197886` `interchainSecurityModule()` = aggregation ISM, `modulesAndThreshold` = 3 modules / threshold 3.

## Test plan
- [ ] Warp Config Checker clears the KII/kiichain `interchainSecurityModule` mismatch on next run.

### DIFF
--- a/.changeset/kii-kiichain-leg-ism-catchup.md
+++ b/.changeset/kii-kiichain-leg-ism-catchup.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Recorded the on-chain interchainSecurityModule for the KII/kiichain warp route's kiichain leg. The kiichain leg was left out of #1666 while the chain was halted, but the router has the same staticAggregationIsm (threshold = all) wrapping a rateLimitedIsm (1M KII/day), a pausableIsm, and a defaultFallbackRoutingIsm as the EVM legs. The rate-limit and fallback modules are owned by the kiichain token owner; the pausable module is owned by the dedicated Turnkey pauser key. This aligns the registry with on-chain state now that kiichain is live.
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
Reviewed exact head 012b6ab7. The Kiichain leg matches the live router 3-of-3 ISM graph: 1M KII per 86400s rate limit with the expected owner, recipient, and mailbox; unpaused dedicated pauser; empty default-fallback routing ISM with the expected owner; and threshold 3. Exact-head format, lint, build, 8,268 unit tests, file validators, changeset status, and diff checks pass. check-warp-deploy config sync also passes; its red on-chain bucket is only Mantle explorer invalid-JSON contract-verification noise, independently triaged and unrelated to this change.

PR #1686, opened 2026-08-31
chore(radix routes): add pausable aggregation ISM to all EVM legs

haggis-hyperlane:
## Summary

Sets every EVM leg of every Radix warp route to a pausable ISM so the Radix route owner can pause inbound delivery on all supported legs in response to the Radix incident. (Started as the USDC/radix ethereum leg; extended here to all legs. #1688 consolidated in.)

Each EVM leg's `interchainSecurityModule` is a 2-of-2 `staticAggregationIsm` of:
- `defaultFallbackRoutingIsm` (owner = route owner `0xA365…EB8F`, empty domains → mailbox default)
- `pausableIsm` (`paused: false`, owner = turnkey pauser `0x60Cc386C85717CB51C2827A75e14826883dF5da4`)

Pausing the pausableIsm makes the aggregation fail closed → halts inbound releases on that leg.

### Legs
- USDC/radix: `arbitrum`, `base`, `ethereum`
- USDT/ethereum-radix: `ethereum`
- XRD/radix: `base`, `ethereum`
- WBTC/ethereum-radix: `ethereum`
- BNB/radix: `bsc`
- ETH/ethereum-radix: `ethereum`

### Not changed (aggregation/pausing not supported)
- All `radix` legs — Radix ISM impl supports only multisig/routing/noop.
- `solanamainnet` legs — pausable/aggregation only exist as nested composite-ISM nodes; pauser is an EVM key.

### Follow-up
Per-chain the routes will share a single aggregation ISM (config may be switched to reference the shared deployed ISM address per chain). Applying requires deploying the shared ISM per chain and the route owner `0xA365…EB8F` (an EOA we don't control) signing `setInterchainSecurityModule` on each router.

Source Haggis thread: https://haggis.services.hyperlane.xyz/threads/01M1CVBBP7EM6KQGCA418N8NPH

### DIFF
--- a/.changeset/radix-routes-pausable-ism-all-legs.md
+++ b/.changeset/radix-routes-pausable-ism-all-legs.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+the EVM legs of all Radix warp routes (USDC/radix arbitrum+base, USDT/ethereum-radix ethereum, XRD/radix base+ethereum, WBTC/ethereum-radix ethereum, BNB/radix bsc, ETH/ethereum-radix ethereum) were set to a 2-of-2 aggregation ISM combining the default fallback ISM with a pausable ISM owned by the turnkey pauser, so every EVM leg can be paused in response to the radix incident wherever aggregation/pausing is supported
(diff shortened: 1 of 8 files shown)


paulbalaji (approved):
Reviewed ddf7ac8e. The 2-of-2 composition preserves mailbox-default verification while unpaused and makes the Turnkey-owned pausable ISM mandatory. Schema/config validation passes; the outstanding on-chain wiring is accurately disclosed.

PR #1691, opened 2026-09-02
chore: update Zod to 4.5.4

paulbalaji:
Bumps Zod from 4.5.2 to 4.5.4.

Upstream fix: https://github.com/colinhacks/zod/pull/6500

Zod 4.5.4 fixes function-valued `.default()`/`.prefault()` factories firing during schema cycle detection. This is a patch-only dependency update.

Changeset: patch bump for `@hyperlane-xyz/registry`.

Verification:
- Frozen lockfile-only install passed supply-chain policy.
- Full install completed successfully.
- Unit suite/build passed.
- Direct 4.5.4 regression script passed: `z.compile()` does not invoke a default factory; supplied values do not invoke it; absent value invokes it once.
- Changeset status passed.

### DIFF
--- a/.changeset/quiet-zebras-compile.md
+++ b/.changeset/quiet-zebras-compile.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Zod was updated to 4.5.4 to prevent function-valued default factories from running during schema cycle detection.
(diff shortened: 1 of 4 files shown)


haggis-hyperlane (approved):
Approved: patch-only Zod bump 4.5.2 → 4.5.4. Diff is package.json + lockfile + changeset only; all CI checks passing.

PR #1710, opened 2026-09-18
fix(arc): add gas transactionOverrides for prompt tx inclusion

haggis-hyperlane:
## Summary
- Adds `transactionOverrides` to `chains/arc/metadata.yaml`: `maxFeePerGas: 60 gwei`, `maxPriorityFeePerGas: 2 gwei`.
- Fixes recurring `Key Funder Pods Failing` / `3+ pod restarts` low-urgency alerts caused by arc funding timeouts.

## Root cause
The mainnet3 key-funder hourly CronJob fails `1/64 chains failed to fund: arc` with `Funding timed out for chain arc`. Investigation showed:
- arc RPC (Alchemy private secret) and the public registry RPC were both **up** with 0 errors during every failure window — not an RPC/downtime problem.
- Funder balance is healthy (~19 ARC); not a balance problem.
- The arc IGP-claim tx (runs first, holds the funder nonce) is **broadcast and accepted** but **not included**: one case was mined a full hour later; two others were dropped from the mempool. No nonce gap — txs used the next valid nonce.
- Txs carried only the ~0.1 gwei suggested priority fee over a steady 20 gwei base fee, which arc's sequencer treats as too low to include promptly.

## Fix
Set an explicit fee override so txs carry a meaningful tip and ample ceiling. arc's native token is USDC, so even at 60 gwei a ~40k-gas claim tx costs a fraction of a cent.

## Test plan
- [ ] Merge + registry release, then confirm the next key-funder runs complete arc funding without `Funding timed out for chain arc`.
- [ ] Verify arc claim/top-up txs are mined within a block or two on-chain.

Source Haggis thread: https://haggis.services.hyperlane.xyz/threads/01M2ST1VXSZ0SCCNB3E7ZHH5FE


troykessler (approved):

.changeset/arc-tx-gas-overrides.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/registry': patch
+---
+
+Arc gained transactionOverrides setting maxFeePerGas to 60 gwei and maxPriorityFeePerGas to 2 gwei so transactions are included promptly. Previously txs used a ~0.1 gwei tip and lingered in the mempool for up to an hour or were dropped, causing key funder IGP-claim/top-up timeouts. Gas is paid in USDC so the higher ceiling costs a fraction of a cent per tx.
```
paulbalaji, line 5: Non-blocking rollout note: please include updating/redeploying the key funder's registry pin in the test plan. `KeyFunderHelmManager.helmValues()` sets `REGISTRY_URI` to `/tree/${registryCommit}`, and the service loads metadata from that URI. A registry merge/release alone will not update a CronJob pinned to the old commit. Verify the deployed URI includes this change before checking the next claim/top-up run.

paulbalaji (approved):
Reviewed d85ab1b6481860bd13c87c2b330977e9678659f8, including downstream fee handling. No blocking findings.

Local registry build, Arc schema validation, and an SDK MultiProvider.prepareTx check passed; the claim and funding paths both use these overrides. CI is green. Live Arc RPC returned chain ID 5042, a 20 gwei base fee, and a 0.1 gwei suggested tip. This supports the configuration choice, but does not prove the incident is resolved; inclusion latency still needs the planned post-deployment check.

Second-order effect: these are chain-wide fixed transaction values, so SDK deployments/transfers also inherit them. Rust agent config generation preserves them, and both the legacy relayer and Lander initial fee estimators replace their dynamic estimates with these values. Lander can still escalate replacements. For SDK sends, a base fee above 60 gwei would prevent inclusion until the override is updated. Reasonable at the observed base fee, but broader than a key-funder-only fix.

